PR #298, opened 2015-01-02
Package Update Format

Jiehong:
This pull request is a new version of https://github.com/aurapm/aura/pull/268.

I've made is shorter, and adapted to the new rst format.


doc/source/design.rst
```diff
@@ -335,10 +335,21 @@ Hooks are passed through Aura as an ADT of functions.
 Aesthetics
 ----------
 
+Size Information
+~~~~~~~~~~~~~~~~
+If ``--verbose-size`` is passed to Aura, the following information are
```
DuncanvR: ... the following information *is* displayed ...

doc/source/design.rst
```diff
@@ -357,7 +368,36 @@ example:
 Version Information When Upgrading
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--  Need a nice chart.
+Whenever a package needs an upgrade, if a ``--verbose-update`` is passed to
```
DuncanvR: Whenever a package needs an upgrade and \`\`--verbose-update\`\` ...

doc/source/design.rst
```diff
@@ -357,7 +368,36 @@ example:
...
+Aura, then a detailed chart is produced, as described below.
+
+The coloured part is denoted with a ``<colour></colour>`` syntax.
```
DuncanvR: ```... with ``<colour></colour>`` tags```, i.e. leave out the `a`.
Additionally, should we give an example of how to use those colour tags here?

DuncanvR:
I'm not familiar with the rst format, but apart from the couple of text-based comments it LGTM!

Jiehong:
@DuncanvR 
> Additionally, should we give an example of how to use those colour tags here?

The colour tags are only here to indicate which part should be coloured, but they would never be print in the output.
The idea is to simply highlight what part of the version number has changed, because it can be a little bit annoying to distinguish sometimes.

Other than that, I am unsure of what example I could give, since I gave an example for each subsection.

DuncanvR:
You're welcome. And happy new year! :)

Do you mean you'd use them like this: `<colour>coloured text here</colour>`? Maybe we could just put that in there then?

Jiehong:
Happy new year!

Yes, I thought about them as xml tags, but only for the documentation. Perhaps I should provide more detail on that tag then.

bb010g:
:+1:
Along with @DuncanvR's comments, LGTM.

doc/source/design.rst
```diff
@@ -335,10 +335,21 @@ Hooks are passed through Aura as an ADT of functions.
 Aesthetics
 ----------
 
+Size Information
+~~~~~~~~~~~~~~~~
+If ``--verbose-size`` is passed to Aura, the following information is
+displayed before installation from the official repositories (may not be
+possible for AUR)
```
fosskers: The `AUR` line isn't needed here, as not all Aura backends with have access to the AUR. Only Arch ones will.

doc/source/design.rst
```diff
@@ -335,10 +335,20 @@ Hooks are passed through Aura as an ADT of functions.
...
 
+Size Information
+~~~~~~~~~~~~~~~~
+If ``--verbose-size`` is passed to Aura, the following information is
+displayed before installation from the official repositories.
+
+.. code-block::
+
+   Total download size: xx MiB
+   Net upgrade size   : xx MiB
```
fosskers: MiB -> MB I think.
Je pense que `MiB` est français, non?

Jiehong: No, MiB is English (https://en.wikipedia.org/wiki/Mebibyte).
You can see it in the output of `ifconfig` for example, or gparted, and so on. Those binary prefixes are meant to replace the conflicting ones with the SI prefixes.

In French, it would be "Mio".

fosskers:
I agree with the format of this. Two things:

1. Should this be the default behaviour? Rather than having to use a `--verbose-foo` flag.
2. I think this could use some Aura styling. Doesn't have to be the current `aura >>=`, but something similar couldn't hurt.

Jiehong:
@fosskers:

1. I don't mind ^^, but is that ok with everyone? Pacman has a similar option (°VerbosePkgLists`), but it is disabled by default;

2. I've used `== `, or maybe you meant something different by "aura styling". Pacman seems to use `::` in blue. We could either use the same, or keep on standing out with something different (or crazy with a unicode arrow such as `→ ` or `⇒`).

fosskers:
+1 for unicode. I like `⇒` as that's actually recognized by Haskell to be equivalent to `>>=`.

I think our default output should be detailed, with the ability to silence it further with `-q` options.

fosskers:
Thanks for this. Everything looks good.

PR #300, opened 2015-01-04
Do not report an error when all requested packages are already installed

### LINKED ISSUE #299: Exit with success when specifying --needed and nothing to do
Aura reports an error (nonzero exit code) when all requested packages are already installed, even if `--needed` is specified. For uniformity with Pacman (and script use), this should not be an error.

Compare:

```bash
$ sudo pacman -S --needed abs; echo $?
warning: abs-2.4.4-2 is up to date -- skipping
 there is nothing to do
0
```

```bash
$ sudo aura -A --needed google-talkplugin; echo $?
aura >>= The following packages are already installed:
google-talkplugin

aura >>= No valid packages specified.
1
```

koterpillar:
- remove maybeScoldAndFail as unused

Fixes #299.


fosskers:
Merged into `aura1` branch. Thanks!

PR #302, opened 2015-01-11
update Portuguese translation

thiagowfx:
Small updates to the portuguese translation.


fosskers:
Thanks! This has been merged into the `aura1` branch.

PR #308, opened 2015-03-11
Update and fix russian translation

exlevan:


fosskers:
Thanks for this! I'll be merging it into the `aura1` branch in the next few days.

fosskers:
Merged into `aura1`.

PR #313, opened 2015-03-25
Update german translation

jplatte:


src/Aura/Languages.hs
```diff
@@ -339,9 +340,10 @@ buildFail_6 = \case
 -- NEEDS TRANSLATION
 buildFail_7 :: String -> Language -> String
 buildFail_7 (bt -> p) = \case
-    Japanese -> p ++ "の作成スクリプトを収得できなかった。"
-    Indonesia-> "Gagal mendapatkan skrip untuk " ++ p ++ "."
-    _        -> "Failed to obtain build scripts for " ++ p ++ "."
+    Japanese  -> p ++ "の作成スクリプトを収得できなかった。"
+    German    -> "Herunterladen der Build-Skripte für " ++ p ++ " fehlgeschlagen."
```
jplatte: 'herunterladen' is German for 'download'. Other programs (e.g. apt-get) use 'holen' instead which is a more direct translation. However, it sounds a lot better to use 'herunterladen' in contexts where 'obtain' is used interchangably with 'download'. Also, 'Build-Skripte' is kind of a semi-translation, but way better than 'Bau-Skripte' (using the verb 'bauen' as translation for 'build' here like in translations before). The latter would sound ridiculous and wouldn't be better understood IMHO – the translation 'bauen' is already a poor one that only exists because Germanizing 'build' ('builden') sounds even worse.

src/Aura/Languages.hs
```diff
@@ -1576,7 +1580,7 @@ singleSync_1 (bt -> p) = \case
 absInfoFields :: Language -> [String]
 absInfoFields = \case
     Croatian   -> [ "Repozitorij","Ime","Verzija","Zavisnosti","Make Zavisnosti","Opis" ]
-    German     -> [ "Quelle","Name","Version","Hängt ab von","Make Abh.","Beschreibung"]
+    German     -> [ "Repository","Name","Version","Hängt ab von","Make-Abhängigkeiten","Beschreibung"]
```
jplatte: 'Quelle' might be a usable translation, but it's rather unclear (would be a direct translation for 'source') and I think 'Repository' is a widely known term among german linux sys-admins.

src/Aura/Languages.hs
```diff
@@ -1630,7 +1634,7 @@ saveState_1 :: Language -> String
 saveState_1 = \case
     Japanese   -> "現在パッケージ状態保存完了。"
     Croatian   -> "Stanje paketa spremljeno."
-    German     -> "Paketzustände gesichert."
+    German     -> "Paketzustand gesichert."
```
jplatte: The old translation was for 'Saved package states.' instead of 'Saved package state.'.

bb010g:
:+1: (Don't speak German, but the formatting fixes are nice and you explain what you're doing well.)

jplatte:
Haha, thanks! :)

fosskers:
Danke sehr! I'm going to merge this into the `aura1` branch on Sunday.

fosskers:
Merged into `aura1`.

PR #314, opened 2015-03-25
Fix indentation

jplatte:
* Replaced the few existing tabs with spaces in aura.cabal
* Replaced the 8-space indentation with 4-space indentation in _aura (first hunk)
* Replaced all the remaning indentation in _aura, which was tabs, by 4 spaces for a tab

There are still two files that use tab indentation: `doc/Makefile` and `doc/make.bat`. The latter also uses CRLF (`\r\n`) like Windows instead of LF (only `\n`) like Unix-like systems. They use tabs (and CRLF) consistently though, so I didn't change any of that as I thought it might be intended.


fosskers:
Danke, nochmal. Like the other PR, this is going in on Sunday.

fosskers:
Merged into `aura1`. Thanks again!

PR #316, opened 2015-04-06
Honor two-word pacman options

Tomsod:
Now the options parser recognises options
like --arch or --config and passes them
to pacman along with their arguments.

I also had to gut some of the
parseLanguageFlag code which was
messing up other options before I even
could get my hands on them.

### DIFF
--- a/src/Aura/Flags.hs
+++ b/src/Aura/Flags.hs
@@ -26,6 +26,7 @@ module Aura.Flags
     , reconvertFlags
     , dualFlagMap
     , hijackedFlagMap
+    , pacmanFlagMap
     , buildABSDepsStatus
     , confirmationStatus
     , customizepkgStatus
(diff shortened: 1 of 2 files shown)


src/Aura/Flags.hs
```diff
@@ -192,6 +196,21 @@ dualOptions = Option [] ["ignore"]      (ReqArg Ignore      "" ) "" :
...
+-- but parser needs to know that they require an argument
+longPacmanOptions :: [OptDescr Flag]
+longPacmanOptions = map pacArg
+                    [ "dbpath", "root", "arch", "cachedir", "color"
+                    , "config", "gpgdir" , "logfile", "assume-installed"
+                    , "print-format" ]
+                    -- "owns" is apparently okay as is?
+                    -- TODO: check all others
+    where pacArg option = Option [] [option] (ReqArg (PacmanArg option) "") ""
+
```
fosskers: Do none of these options have short versions?

Tomsod:
Right, good catch! I completely missed that.
And I should give a warning:
I was not able to test-compile this with new ghc:
it showers me with errors about new compile flag.
This might prove being troublesome.

fosskers:
#318 has been fixed, so you can test with the latest version on the `aura1` branch.

fosskers:
Thanks for this!

PR #327, opened 2015-05-26
Fix build on GHC 7.10

danr:
GHC 7.10 is more picky about that inferred types are actually allowed under the current set of extensions, so FlexibleContexts is needed in a few places. Also, Data.Time now exports a months function.

### DIFF
--- a/src/Aura/Commands/A.hs
+++ b/src/Aura/Commands/A.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE FlexibleContexts #-}
 
 -- Handles all `-A` operations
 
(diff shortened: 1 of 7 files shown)


adituv:
All instances of FlexibleContexts may be replaced by monomorphic type signatures if the maintainers would prefer that.  I have made my own fork with the same changes under adituv/aura/710patch

adituv:
FlexibleContexts may be replaced by monomorphic type variables in some cases; I have created my own fork with the relevant changes here: https://github.com/danr/aura/compare/master...adituv:710patch

This can be just a matter of taste for the maintainers; personally, I prefer just adding FlexibleContexts as I find having just occasional type signatures for where-bindings a little ugly, but I'm submitting this as an option for the maintainers.

fosskers:
Hey guys, this has already been fixed on another branch, which I've merged into master.

PR #337, opened 2015-08-07
add stack.yaml

joehillen:


fosskers:
Have you built aura with stack? If so, what's the experience like?

joehillen:
@fosskers Yes, [stack](https://github.com/commercialhaskell/stack) is 100x better than `cabal-install`. Try it out. I'll probably never use `cabal-install` ever again.

joehillen:
Still waiting on travis to add support for stack. I have no idea what the hold up is. https://github.com/travis-ci/apt-package-whitelist/issues/379

fosskers:
What's been your experience using it, more specifically? I'd like more 1st-hand accounts from happy people before I attempt a switch.

joehillen:
You can use both side-by-side. They don't conflict.

```
sudo aura -A haskell-stack
stack setup # only need to run once to download ghc
stack build
```
beyond that is has a ton of awesome features. `cabal-install` has always been a giant mess and it's not worth clinging to.

bb010g:
:+1: for building with `stack`. Using it for a new project and it's working great.

fosskers:
Which GHC is it? We need at least 7.10.

joehillen:
[lts-3.1](https://www.stackage.org/lts-3.1) is 7.10.2

fosskers:
@bb010g So long as I can get `haskell-mode` to work with stack, I'd be willing to switch. That said, I don't actually mind sandboxes, except having unnecessary duplicates of large libraries sometimes.

joehillen:
@fosskers Here is my [haskell-mode-init.el](https://gist.github.com/joehillen/47899db33bdf632a9788) if that helps.

fosskers:
Thanks, I'll see if I can get that going.

PR #338, opened 2015-08-07
Add --noconfirm to pacman -U

joehillen:
Aura already asks if we want to install the packages.
There is no need to ask again after makepkg.


fosskers:
If `--noconfirm` is given as an aura flag, will this conflict at the pacman level?

joehillen:
Turns out I don't actually want this behavior. Sometimes pacman will ask questions other than the usual "Install? Upgrade?" like "Replace foo with bar?" where `--noconfirm` is bad behavior.

fosskers:
Fair enough.

PR #342, opened 2015-08-11
Add Localized Yes-/No-Prompt

markusgrosser:
Translations have been taken from the [pacman repository](https://projects.archlinux.org/pacman.git/tree/src/pacman/po).

This also slightly changes behaviour for the English version, since the regex has changed to mimic pacman's behaviour.

### DIFF
--- a/src/Aura/Languages.hs
+++ b/src/Aura/Languages.hs
@@ -1785,7 +1785,7 @@ yesNoMessage = \case
     _          -> "[Y/n]"
 
 yesRegex :: Language -> String
-yesRegex = \case
+yesRegex = (++"|^$") . \case
     Croatian   -> "[dD][aA]?"
     German     -> "[jJ][aA]?"
     Norwegian  -> "[jJ][aA]?"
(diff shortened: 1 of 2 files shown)


fosskers:
Meaning pacman will accept `[yY]([eE][sS])`?

markusgrosser:
That's what I took from [pacman's code](https://projects.archlinux.org/pacman.git/tree/src/pacman/util.c#n1580); the `_(...)` macro substitutes with the translated string:

```c
if(strcasecmp(response, _("Y")) == 0 || strcasecmp(response, _("YES")) == 0) {
	return 1;
} else if(strcasecmp(response, _("N")) == 0 || strcasecmp(response, _("NO")) == 0) {
	return 0;
}
```

I do agree that the Regex looks a bit ugly and could probably be prettied up by caseless comparison, but I don't quite understand how that works.

fosskers:
Meaning Aura would accept `YeS`, but I don't see that as a problem.

fosskers:
Will these new Regexes accept just pressing the enter button, without any input? Notice the `\\B` in the original code.

markusgrosser:
Oh, seems I missed that it matches empty strings. Reason was that `\\B` causes any input of more than two letters to be passed (so `no` was interpreted as a yes) and thus didn't include it. It now uses matching again `^$` to accept empty strings as well.

fosskers:
Vielen Dank, Markus.

PR #343, opened 2015-08-11
Fix Language option parsing

markusgrosser:
Before, any language option other than --english caused Aura to crash
with an "incompatible options" error message.

This was caused by the language option being handed to the option
parser as an additional flag, causing all case  branches to be dropped
through.

This bug seems to have gone unnoticed, I discovered it while trying to test #342


fosskers:
This has been fixed.

PR #346, opened 2015-08-22
Enable downloading build files using git

markusgrosser:
This adds a flag that enables downloading of build files (PKGBUILD, .install etc.) using git, which was made possible with AUR4.

While this currently doesn't give any benefit over the old method of downloading a tarball*, it is intended to serve as a baseline for possible future changes, like using git for diffs.

It also adds [hlibgit2](http://hackage.haskell.org/package/hlibgit2) as a dependency, which currently doesn't have an AUR package. I'm willing to create and maintain one (as well as dependencies), should this get merged.

As a useless side effect, this makes aura the first AUR Helper to have all features in the [AUR Helper comparison table](https://wiki.archlinux.org/index.php/AUR_helpers#Comparison_table).

&nbsp;

*In fact, this just downloads more data since it needs to fetch the complete history.

### DIFF
--- a/aura.cabal
+++ b/aura.cabal
@@ -96,6 +96,7 @@ executable aura
                        parsec,
                        process,
                        filepath,
+                       hlibgit2,
                        directory,
                        temporary,
                        containers,
--- a/doc/aura.8
+++ b/doc/aura.8
@@ -85,11 +85,6 @@ packages, so all dependencies of dependencies (and so on) will also be shown.
 This can aid the pre-install package research process.
 .RE
 .P
-\fB\-g\fR, \-\-use-git
-.RS 4
-Use \fIgit clone\fR to download build files instead of a tarball.
-.RE
-.P
 \fB\-i\fR, \-\-info <package(s)>
 .RS 4
 View information about a package in the AUR.
--- a/doc/aura.8
+++ b/doc/aura.8
@@ -85,6 +85,11 @@ packages, so all dependencies of dependencies (and so on) will also be shown.
 This can aid the pre-install package research process.
 .RE
 .P
+\fB\-g\fR, \-\-use-git
+.RS 4
+Use \fIgit clone\fR to download build files instead of a tarball.
+.RE
+.P
 \fB\-i\fR, \-\-info <package(s)>
 .RS 4
 View information about a package in the AUR.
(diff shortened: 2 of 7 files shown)


fosskers:
If we're going to do this, let's just make it the default.

flamusdiu:
@fosskers I would have to agree. All the AUR packages are going to be in a git repo anyway. It would support better downloading too.

fosskers:
How does this affect `-Aw`? Will it just clone the repo to the `pwd`?

fosskers:
@markusgrosser , I just checked Hackage and there are quite a few git packages. What was the motivation for choosing the one you did?

markusgrosser:
The main reasons were feature-completeness and documentation, in which [gitlib](http://hackage.haskell.org/package/gitlib) was (and still is) sadly lacking.

[libgit](http://hackage.haskell.org/package/libgit) (not to be confused with libgit2 which is the C library) might be an alternative, but it's only a wrapper around the git executable, whereas I personally preferred a "native" implementation.

I didn't see [hit](http://hackage.haskell.org/package/hit) when I initially searched for a git library, but it seems like a good alternative, which I'll definitely look into. It doesn't support cloning though, so that'll probably have to be done by calling `git clone ...` directly.

Regarding `-Aw`, it seems to not do … anything, which I can't say I really understand. Cloning into `$PDW/$pkgname` seems like a sane choice though, so I'll probably implement that.

fosskers:
> git

`gitlib` was the main one I looked at, but was fairly quickly turned off by it. The main things we're looking for are cloning and the ability to branch, to be able to hotedit PKGBUILDs.

> Cloning into $PDW/$pkgname seems like a sane choice

I agree, that's the end result of what `-Aw` does, after you've untar'd the file it downloads.

DuncanvR:
@markusgrosser
> *In fact, this just downloads more data since it needs to fetch the complete history.

Not sure how this is achieved using hlibgit2, but on the command line you can use the [--depth](https://www.git-scm.com/docs/git-clone#git-clone---depthltdepthgt) argument to control the amount of history that's fetched. E.g. `git clone --depth 1 https://github.com/aurapm/aura.git` only retrieves the last commit -- but still with all files in place.

AladW:
Impressive work! The wiki entry for aura should probably be updated...

AladW:
Or I'll wait until the AUR pkgver is bumped: https://github.com/aurapm/aura/issues/489

fosskers:
The README has been updated, but both Aura's wiki entry on the Arch wiki and the AUR helpers chart need to be updated.

PR #364, opened 2015-10-19
Bash parser now parses expansions and the simplifier simplifies them

### LINKED ISSUE #328: Dependency package not correctly parsed
How to reproduce:

``` shell
$ sudo aura -A retroshare
aura >>= Determining dependencies...
aura >>= Dependency checking failed for these reasons:
The dependency `qt4[@]}[@]}[@]}` could not be found. You may need to search for a package to satisfy it.
```

However, the PKGBUILD of the package mentions `qt4` (https://aur.archlinux.org/packages/re/retroshare/PKGBUILD) but not the weird output Auras complains about.

theotherjimmy:
Should fix #328

### DIFF
--- a/src/Bash/Base.hs
+++ b/src/Bash/Base.hs
@@ -22,6 +22,7 @@ along with Aura.  If not, see <http://www.gnu.org/licenses/>.
 module Bash.Base where
 
 import qualified Data.Map.Lazy as M
+import Data.Either
 
 ---
 
@@ -78,23 +79,14 @@ toNamespace (_:fs) = toNamespace fs
 getVar :: Namespace -> String -> Maybe [String]
 getVar ns s = case M.lookup s ns of
   Nothing -> Nothing
-  Just bs -> Just $ foldr (++) [] $ map fromBashString bs
+  Just bs -> Just $ foldMap fromBashString bs
 
 fromBashString :: BashString -> [String]
 fromBashString (SingleQ s) = [s]
-fromBashString (DoubleQ l) = fromBashStringList l
-fromBashString (NoQuote l) = fromBashStringList l
+fromBashString (DoubleQ l) = rights l
+fromBashString (NoQuote l) = rights l
 fromBashString (Backtic c) = ['`' : unwords (fromCommand c) ++ "`"]
 
-fromBashStringList :: [Either BashExpansion String] -> [String]
-fromBashStringList l = map (\e -> case e of
-                                    Left _ -> error "I thought I filtered this out"
-                                    Right s -> s
-                                    )
-                       $ filter (\e -> case e of
-                                   Left _ -> False
-                                   Right _ -> True) l
-
 fromCommand :: Field -> [String]
-fromCommand (Command c as) =  [c] ++ foldr (++) [] (map fromBashString as)
+fromCommand (Command c as) =  c : foldMap fromBashString as
 fromCommand _ = error "Argument given was not a Command."
--- a/src/Bash/Base.hs
+++ b/src/Bash/Base.hs
@@ -52,10 +52,13 @@ data BashFor = Incr  -- for (x;y;z); do ... done  -- Incomplete!
 -- | While `String` is the main data type in Bash, there are four
 -- subtypes each with different behaviour.
 data BashString = SingleQ String
-                | DoubleQ String
-                | NoQuote String
+                | DoubleQ [Either BashExpansion String]
+                | NoQuote [Either BashExpansion String]
                 | Backtic Field   -- Contains a Command.
                   deriving (Eq,Show)
+
+data BashExpansion = BashExpansion String [BashString]
+                     deriving (Eq,Show)
 
 type Namespace = M.Map String [BashString]
 type Script    = [Field]  -- A parsed Bash script.
@@ -74,15 +77,24 @@ toNamespace (_:fs) = toNamespace fs
 -- | Never call this directly. Use `value` in `Aura/Bash`.
 getVar :: Namespace -> String -> Maybe [String]
 getVar ns s = case M.lookup s ns of
-                Nothing -> Nothing
-                Just bs -> Just $ map fromBashString bs
+  Nothing -> Nothing
+  Just bs -> Just $ foldr (++) [] $ map fromBashString bs
 
-fromBashString :: BashString -> String
-fromBashString (SingleQ s) = s
-fromBashString (DoubleQ s) = s
-fromBashString (NoQuote s) = s
-fromBashString (Backtic c) = '`' : fromCommand c ++ "`"
+fromBashString :: BashString -> [String]
+fromBashString (SingleQ s) = [s]
+fromBashString (DoubleQ l) = fromBashStringList l
+fromBashString (NoQuote l) = fromBashStringList l
+fromBashString (Backtic c) = ['`' : unwords (fromCommand c) ++ "`"]
 
-fromCommand :: Field -> String
-fromCommand (Command c as) = unwords $ c : map fromBashString as
+fromBashStringList :: [Either BashExpansion String] -> [String]
+fromBashStringList l = map (\e -> case e of
+                                    Left _ -> error "I thought I filtered this out"
+                                    Right s -> s
+                                    )
+                       $ filter (\e -> case e of
+                                   Left _ -> False
+                                   Right _ -> True) l
+
+fromCommand :: Field -> [String]
+fromCommand (Command c as) =  [c] ++ foldr (++) [] (map fromBashString as)
 fromCommand _ = error "Argument given was not a Command."
--- a/src/Bash/Parser.hs
+++ b/src/Bash/Parser.hs
@@ -104,7 +104,9 @@ singleQuoted = between (char '\'') (char '\'')
 -- | Replaces ${...}. No string extrapolation.
 doubleQuoted :: Parser [BashString]
 doubleQuoted = between (char '"') (char '"')
-               ((\s -> [DoubleQ s]) <$> many1 (noneOf ['\n','"']))
+               ((\s -> [DoubleQ s]) <$> many1 (choice [ try (Left <$> expansion)
+                                                      , Right <$> many1 (noneOf ['\n','"','$'])
+                                                      ]))
                <?> "double quoted string"
 
 -- | Contains commands.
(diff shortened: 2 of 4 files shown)


src/Bash/Base.hs
```diff
@@ -74,15 +77,24 @@ toNamespace (_:fs) = toNamespace fs
...
+  Nothing -> Nothing
+  Just bs -> Just $ foldr (++) [] $ map fromBashString bs
+
+fromBashString :: BashString -> [String]
+fromBashString (SingleQ s) = [s]
+fromBashString (DoubleQ l) = fromBashStringList l
+fromBashString (NoQuote l) = fromBashStringList l
+fromBashString (Backtic c) = ['`' : unwords (fromCommand c) ++ "`"]
+
+fromBashStringList :: [Either BashExpansion String] -> [String]
```
fosskers:
```
fromBashStringList :: [Either BashExpansion String] -> [String]
fromBashStringList = rights
```
This would do the job.

theotherjimmy: Thanks! I'll update my PR.

fosskers: You can probably just get rid of that function altogether, and just call `rights` directly.

theotherjimmy: I did :] thanks.

src/Bash/Base.hs
```diff
@@ -74,15 +78,15 @@ toNamespace (_:fs) = toNamespace fs
 -- | Never call this directly. Use `value` in `Aura/Bash`.
 getVar :: Namespace -> String -> Maybe [String]
 getVar ns s = case M.lookup s ns of
-                Nothing -> Nothing
-                Just bs -> Just $ map fromBashString bs
+  Nothing -> Nothing
+  Just bs -> Just $ foldr (++) [] $ map fromBashString bs
```
bb010g: `foldr (++) []` can just be `map fold`. (Also, @fosskers, what's aura style with regards to fmap vs. map?)

src/Bash/Base.hs
```diff
@@ -74,15 +78,15 @@ toNamespace (_:fs) = toNamespace fs
...
+fromBashString :: BashString -> [String]
+fromBashString (SingleQ s) = [s]
+fromBashString (DoubleQ l) = rights l
+fromBashString (NoQuote l) = rights l
+fromBashString (Backtic c) = ['`' : unwords (fromCommand c) ++ "`"]
 
-fromCommand :: Field -> String
-fromCommand (Command c as) = unwords $ c : map fromBashString as
+fromCommand :: Field -> [String]
+fromCommand (Command c as) =  [c] ++ foldr (++) [] (map fromBashString as)
```
bb010g: `[c] ++ map fold (map fromBashString as) === c : map (fold . fromBashString) as`

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
 
 replaceStr :: BashString -> State Namespace BashString
 replaceStr s@(SingleQ _) = return s
-replaceStr (DoubleQ s)   = DoubleQ `fmap` replaceStr' s
-replaceStr (NoQuote s)   = NoQuote `fmap` replaceStr' s
+replaceStr (DoubleQ s)   = expandStr s >>= \strs -> return $ DoubleQ $ map Right strs
```
bb010g: `expandStr s >>= (return . DoubleQ . map Right)`
Another style question, @fosskers: `return` or `pure`? I'd imagine `pure` due to AMP.

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
...
+  lhs <- expandStr' x
+  rhs <- expandStr xs
+  return $ lhs ++ rhs
+
+expandStr' :: Either BashExpansion String -> State Namespace [String]
+expandStr' (Right s) = return [s]
+expandStr' (Left (BashExpansion var sub)) = do
+  subscript <- mapM expandWrapper sub
+  ns <- get
+  getIndexedString ns var $ foldr (++) "" subscript
```
bb010g: `foldr (++) "" subscript === fold subscript`

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
...
+  Nothing -> return []
+  Just bstrs -> getIndexedString' bstrs idx
+
+getIndexedString' :: [BashString] -> String -> State Namespace [String]
+getIndexedString' bstrs "" = do
+  singlestr <- replaceStr $ head bstrs
+  return $ fromBashString singlestr
+getIndexedString' bstrs "*" = do
+  strings <- mapM replaceStr bstrs
+  return $ foldr (++) [] $ map fromBashString strings
```
bb010g: `return $ map (fold . fromBashString) strings`

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
...
+getIndexedString' bstrs "@" = getIndexedString' bstrs "*"
+getIndexedString' bstrs num = do
+  singlestr <- replaceStr $ bstrs !! read num
+  return $ fromBashString singlestr
+
+expandWrapper :: BashString -> State Namespace String
+expandWrapper (SingleQ s) = return s
+expandWrapper (DoubleQ s) = do
+  strs <- expandStr s
+  return $ foldr (++) "" strs
```
bb010g: `return $ fold strs`

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
...
+expandWrapper (NoQuote s) = do
+  strs <- expandStr s
+  return $ foldr (++) "" strs
```
bb010g: `return $ fold strs`

src/Bash/Simplify.hs
```diff
@@ -74,18 +74,50 @@ replace' somethingElse = return somethingElse
...
+expandWrapper (Backtic f) = replace' f >>= \x -> return $ unwords $ fromCommand x
```
bb010g: `replace' f >>= (return . unwords . fromCommand)`

bb010g:
Related to parsing, do we want to switch over to [Megaparsec](https://hackage.haskell.org/package/megaparsec) ([rationale](https://notehub.org/w7037)).

src/Bash/Simplify.hs
theotherjimmy: I'm 60% sure that should be fold $ map fromBashString strings. Otherwise
you concatenate expanded bash arrays into one element of the bigger array
instead of extending it by the contents of the expanding variable (the
primary use case for this part of bash by my non-scientific survey).

bb010g: Sorry. `concatMap`?

src/Bash/Simplify.hs
theotherjimmy: A good number of the do-style blocks can be reduced to this style. All
cases of expandWrapper and getIndexedString in fact. They are much cleaner.
I will be updating the PR when I wake up tomorrow.

bb010g: If you want to do style adjustments outside of the scope of parsing bash expansions, please make a new PR.

src/Bash/Simplify.hs
theotherjimmy: Yup. Replaced all instances. The code is looking much better. Thanks.

src/Bash/Simplify.hs
theotherjimmy: I'll undo that. It slipped by me.

theotherjimmy: Actually, I'm now confused. Do you mean the style suggestions you were
making or just the indentation things that I missed? Both?

bb010g: I meant for the project as a whole. I thought you were thinking of going through *everything*. Sorry. :) A separate PR would work better there because it's a separate discussion. I'm not quite sure what general aura policy on when you switch to do-notation is, and I'm currently on mobile and not able to check. I think it would be worth a discussion, however. What you may want to do is just use do-notation for now and then PR the style change with this as a sample. As for indentation problems, fixes are always good.

theotherjimmy: Ah. I'm just intent on tidying up my current pull request. I'm only
changing the functions that I made/modified. Nothing so broad. I'll also
take a look (I'm also on mobile), though with the revisions, there is much
less use of do notation. I will make the do to no-do revisions as a sperate
PR, with a discussion on style there as you recommended.

src/Bash/Base.hs
fosskers: `map` is one less character and should always be used for lists.

src/Bash/Simplify.hs
fosskers: `pure`, given that `return` is likely to be removed soon.

fosskers:
@bb010g I like it. I'll start by switching my [new version number library](https://github.com/fosskers/haskell-versions) to it to try it out.

bb010g:
@fosskers On map vs. fmap: The extra character doesn't make a noticeable difference. It's one character, and we have the room. I'd be more prone to use it because of the same reason FTP happened: the type better describes what is needed for the function to work, and it will just keep working if we change data structures. Haskell is not normally about constricting usage where it doesn't need to be constricted.

fosskers:
`Text` only with Aura2. I agree that `fmap` does make it clear that "I'm operating on a Functor now" instead of being constrained to the idea of Lists.

fosskers:
Hey @theotherjimmy , if you make the other style changes as noted in the comments, I will merge this in.

src/Bash/Simplify.hs
bb010g: This can actually just be refactored to `expandWrapper (Backtic f) = unwords . fromCommand <$> replace' f` via `replace' f >>= (return . unwords . fromCommand)` and the monad laws.

theotherjimmy:
I have been working on another branch to do the cleanup. I will merge it (it's a fast-forward I think).

src/Bash/Simplify.hs
theotherjimmy: That is not the only instance. Commits comming soon.

theotherjimmy:
Sorry for a huge pile of commits. They should all correspond to a single style suggestion applied across all of the code I added.

bb010g:
Just squash them.

bb010g:
`mapM` -> `traverse` and `map` -> `fmap`, then LGTM. (`<$>` is used everywhere and it's just `fmap` under the hood.)

theotherjimmy:
I'll add that commit and squash (will rewrite history).

theotherjimmy:
`map` -> `fmap` in the parser or simplifier or both?

bb010g:
Both.

theotherjimmy:
good catch. Updated.

src/Bash/Parser.hs
```diff
@@ -123,10 +137,10 @@ unQuoted = map NoQuote <$> extrapolated []
 -- will not be expanded and will retain their braces.
 extrapolated :: [Char] -> Parser [String]
 extrapolated stops = do
-  xs <- plain <|> bracePair
+  xs <- plain <|>  bracePair
   ys <- option [""] $ try (extrapolated stops)
   return [ x ++ y | x <- xs, y <- ys ]
-      where plain = (: []) `fmap` many1 (noneOf $ " \n{}()" ++ stops)
+      where plain = (: []) `fmap` many1 (noneOf $ " $\n{}[]()" ++ stops)
```
bb010g: `fmap` > `<$>`

src/Bash/Simplify.hs
```diff
@@ -93,7 +113,7 @@ replaceIf i@(If comp fs el) = do
   let (op, l, r) = deComp comp
   left  <- fromBashString `fmap` replaceStr l
   right <- fromBashString `fmap` replaceStr r
```
bb010g: `<$>`

theotherjimmy: I did not touch these lines. Is this really appropliate for this PR?

bb010g: Oh, sorry. *derp*
I'll put in a separate PR for those.

fosskers:
This works.

fosskers:
Thanks for this.

PR #371, opened 2015-11-05
Fix style & genericness

bb010g:
Probably what will end up being early stages right now, but a PR to just generally clean up aura.
* [x] `map` → `fmap` (and `<$>` where appropriate)
* [x] `return` → `pure`
* [x] `>>` → `*>` (and `$>` where appropriate)
* [x] `(a,b)` → `(a, b)`
* [x] List funs → Foldable/Traversable funs
* [x] Monad funs → Functor/Applicative funs
* [ ] Generic types (e.g. `Monad m => m ()` where it really could just be a free `m a`)
* [ ] Imports

### DIFF
--- a/src/Aura/Packages/ABS.hs
+++ b/src/Aura/Packages/ABS.hs
@@ -88,9 +88,9 @@ makeBuildable :: String -> String -> Aura Buildable
 makeBuildable repo name = do
   pb <- absPkgbuild repo name
   pure Buildable { baseNameOf   = name
-                   , pkgbuildOf   = pb
-                   , isExplicit   = False
-                   , buildScripts = \fp -> Just <$> copyTo repo name fp }
+                 , pkgbuildOf   = pb
+                 , isExplicit   = False
+                 , buildScripts = \fp -> Just <$> copyTo repo name fp }
 
 copyTo :: String -> String -> FilePath -> IO FilePath
 copyTo repo name fp = do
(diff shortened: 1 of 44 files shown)


src/Aura/Packages/ABS.hs
```diff
@@ -87,21 +87,21 @@ absDepsRepo = asks (getRepo . buildABSDeps)
 makeBuildable :: String -> String -> Aura Buildable
 makeBuildable repo name = do
   pb <- absPkgbuild repo name
-  return Buildable { baseNameOf   = name
+  pure Buildable { baseNameOf   = name
                    , pkgbuildOf   = pb
                    , isExplicit   = False
                    , buildScripts = \fp -> Just <$> copyTo repo name fp }
```
fosskers: Fields here need to be aligned.

src/Bash/Simplify.hs
```diff
@@ -60,31 +61,31 @@ simpNamespace :: Namespace -> Script -> Namespace
...
-replaceStr (Backtic f)   = Backtic `fmap` replace' f
+replaceStr (Backtic f)   = Backtic <$> replace' f
 
 expandStr :: [Either BashExpansion String] -> State Namespace [String]
 expandStr [] = pure []
 expandStr (x:xs) = do
   lhs <- expandStr' x
   rhs <- expandStr xs
-  pure $ lhs ++ rhs
+  pure $ lhs <> rhs
```
fosskers:
```haskell
expandStr (x:xs) = (<>) <$> expandStr' x <*> expandStr xs
```

fosskers:
Poke :)

bb010g:
Yeah. I'll push what I've got and then stick new stuff in a new PR.

fosskers:
Thanks!

Issue #357, opened 2015-10-05
Respect TTY presence in 'aura --version'


RomanSaveljev:
This is merely a hint on how to comply with a typical adopted practice better. I have noticed calling `aura --version` draws a nice screen animation. Then immediately I tried:

```
aura --version | tee /tmp/f.log
```

The resulting `/tmp/f.log` is full of screen codes. If, for example, we look at `wget` (which uses terminal codes as well), then it handles it more gracefully:

```
$ wget http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux/puppy-tahr/iso/tahrpup%20-6.0-CE/tahr-6.0-CE_noPAE.iso 2>&1 | tee /tmp/wget.log
$ head /tmp/wget.log
--2015-10-05 17:58:20--  http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux/puppy-tahr/iso/tahrpup%20-6.0-CE/tahr-6.0-CE_noPAE.iso
Resolving ftp.nluug.nl (ftp.nluug.nl)... 192.87.102.43, 192.87.102.42, 2001:610:1:80aa:192:87:102:42, ...
Connecting to ftp.nluug.nl (ftp.nluug.nl)|192.87.102.43|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 208666624 (199M) [application/octet-stream]
Saving to: ‘tahr-6.0-CE_noPAE.iso.1’

     0K .......... .......... .......... .......... ..........  0%  393K 8m38s
    50K .......... .......... .......... .......... ..........  0%  389K 8m41s
   100K .......... .......... .......... .......... ..........  0%  753K 7m17s
```

Again, this is more like a cosmetic issue

fosskers:
Similar issue when piping coloured Aura output. This has come up before in issue #102. It'll be fixed trivially in Aura2, where we'll be employing `extensible-effects` to have runtime evironment information available to all the text colouring functions.

fosskers:
Fixed!

Issue #174, opened 2013-12-14
googleearth PKGBUILD parse problems on 32-bit Arch


hutou:
Trying to install GoogleEarth on my 32bits Arch system with 
```
$ sudo aura -A google-earth
```
gives me the following error
```
aura >>= Determining dependencies...
aura >>= Dependency checking failed for these reasons:
The dependency `lib32-fontconfig` could not be found. You may need to search for a package to satisfy it.
```
aura or PKGBUILD culprit ?

Package googleearth correctly installed with 
```
sudo aura -A ld-lsb
yaourt -S google-earth
```
Separate install of ld-lsb was needed because, on first try, "yaourt -S google-earth" complained about lib32-glibc, dependency of ld-lsb

See comments on https://aur.archlinux.org/packages/google-earth/

fosskers:
Fixed.

Issue #102, opened 2013-05-15
Colours should only be shown when the output is a terminal


wilfriedvanasten:
I tend to use grep/less and all kinds of other commands on pacman's search quite a lot. Now I wanted to do this on aura -As search to, but the output is just not that nice because of the escape sequences that would control coloring on terminals:

A sample of output piped into less:

ESC[0;35maur/ESC[0mESC[mESC[1mgtk2hs-ESC[0mESC[1;36mdarcsESC[0m ESC[0;32m20100205-1ESC[0m (ESC[0;33m0ESC[0m)
    A GTK+2 binding for Haskell

This can be avoided by only outputting this if the stdout handle is a terminal. I have to admit I have no idea how to do this in haskell, but I know it is possible. An alternative would be to not use escape sequences but a tty API to control coloring.

v1kn:
I have just encountered this issue too. Unfortunately, while piping `-As` command through `less -R` produces a clean output, redirecting it to a file renders the output unintelligible. I compared this to other colored AUR helpers (`yaourt`, `burgaur`), they do not have this issue.  

Seeing `Aria2` tag, I understand that it will not be implemented in current branch. Is there a TBR date for `Aria2`? It is still the best AUR helper, and I'd love to see this issue fixed.  

Thank You

fosskers:
This will actually be fixed with aura 1.4, which is currently underway.

fosskers:
This should no longer be a problem on `master`, which will be released soon as Aura 2.0.

PR #377, opened 2015-11-19
Aura 1.4

### LINKED ISSUE #158: support pacman’s --color
so far aura doesn’t support `--color=auto` (see #102), which wouldn’t be bad if we could do `--color=never` in scripts.

but aura doesn’t support it at all: `--color=never` makes it fail to output anything, and `--color=never` leaves color on.

dixing this is necessary to fix #106 the way i wanted to do it.

### LINKED ISSUE #331: Cannot get help on individual operations
The default --help message says to use --help or -h with an operation to get help on that specific operation. When I try this, I get this error: 

```
$ aura --help -A
aura >>= Conflicting flags given!
```

fosskers:
# Roadmap to Aura 1.4
- [ ] All shell interaction is with Shelly
- [x] Use `basic-prelude`
- [x] Use `microlens` ~~for aeson interaction~~ (no, just for using `wreq` briefly)
- [x] Let's join the present day. All `String` are now `Text`.
- [ ] Git-based packages
  - [ ] Choose a git package from Hackage (there are many)
  - [ ] Package downloads are git-based, and not tarball (no more `Internet.hs`)
- [ ] Parse `.SRCINFO` for dep information, #174
  - [ ] Write parser
  - [ ] Support `depends_x86_64` etc
  - [ ] Rip out Bash parser (you served us well)
  - [ ] Unit tests for parser (use `tasty`)
  - [ ] Support `epoch`
  - [ ] Support non-integer `pkgrel`
- [ ] CLI flag parsing
  - [ ] Uses `optparse-applicative`. Should fix #331 
  -  [ ] No more "two-stage" option parsing with language flags
- [ ] Use `versions` for parsing and comparing version numbers
- [ ] `extensible-effects` conversion
  - [x] No more `Aura` Monad - use `Effects` type alias for common types
  - [ ] Colouring functions are in `Reader` effect, to fix #158, #102, #357
- [x] Convert Travis config to use `stack` (see `microlens-aeson` for an example)
- [ ] Use the ghc flag `-split-objs` for smaller binary size (but longer compile time)
- [ ] Have `powerpill` use off by default. `--pp` to turn it on.


fosskers:
Note: I haven't checked off the `shelly` item, as there are some further cleanups I'd like to do.

PR #383, opened 2016-01-05
[aura-1.4] Prelude ==> BasicPrelude && Formatting ==> th-printf

theotherjimmy:
This one is much simpler, just making the switch to basic-prelude and changing all of the `sformat`s to a more C-printf looking function. The reasoning of the Formatting to th-printf switch is simple: I did not see th-printf earlier, or I would have used it instead of Formatting from the beginning. My bad XD.

- [x] Prelude ==> BasicPrelude 
- [x] Formatting ==> th-printf

### DIFF
--- a/aura.cabal
+++ b/aura.cabal
@@ -44,6 +44,8 @@ executable aura
   default-extensions:  Rank2Types
                        FlexibleContexts
                        OverloadedStrings
+                       NoImplicitPrelude
+                       QuasiQuotes
 
   default-language:    Haskell2010
 
@@ -94,12 +96,12 @@ executable aura
                        array,
                        attoparsec,
                        base >= 4.8 && < 5,
+                       basic-prelude,
                        bytestring,
                        containers,
                        directory,
                        extensible-effects,
                        filepath,
-                       formatting,
                        lens,
                        megaparsec,
                        process,
@@ -111,6 +113,7 @@ executable aura
                        temporary,
                        text,
                        text-icu,
+                       th-printf,
                        ini,
                        time,
                        transformers,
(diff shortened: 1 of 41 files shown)


bb010g:
Shouldn't BasicPrelude be the first import?

fosskers:
I could only see that mattering if doing so rendered other imports redundant.

bb010g:
It's mainly a style thing. The prelude isn't a "normal" import, and is less noise when set apart from the other imports.

fosskers:
I can agree with that. Let's add that to the style guide (which also has to be updated to reflect the removal of the Aura monad).

theotherjimmy:
Right. Just a moment, I'll update the PR.

On Wed, Jan 6, 2016 at 3:10 PM, Colin Woodbury <notifications@github.com>
wrote:

> I can agree with that. Let's add that to the style guide (which also has
> to be updated to reflect the removal of the Aura monad).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aurapm/aura/pull/383#issuecomment-169461858>.
>



-- 
"Open-source software has fewer bugs because it admits the possibility of
bugs." Paul Graham

fosskers:
Re: `th-printf`. What's its advantage over plain, old `printf`?

flamusdiu:
@fosskers I guess the package is from here: https://github.com/pikajude/th-printf 

Looks like `th-printf` is more flexible then plain `printf` according to the github.

fosskers:
Just took a look. I realize I should probably be using that for everything, haha. Things like:

```haskell
foo name age = "Hello " ++ name ", you are " ++ show age ++ " years old!"
```

are clumsy. I especially like its ability to produce `Text`. I had forgotten normal `printf` was limited in that way.

fosskers:
Thanks again @theotherjimmy for this!

flamusdiu:
@fosskers : lol. The same can be said for python. I guess plenty of language prefer using printf style output in some form for everything.

fosskers:
That shows me there's something inherently good about `printf` style.

PR #400, opened 2016-03-09
started moving to optparse-applicative

theotherjimmy:
WIP.
Questions:
- At the moment, every command line possible parses successfully, leaving aura.hs to interperate the 'pacOpts' flags. Should we instead parse all of pacman's options and allow the parser to fail when some flag matches neither pacman options nor aura options?

More to come as I think of them.
@fosskers, could you comment on the question above?


theotherjimmy:
Oh, I just remembered, I don't think this is supposed to work this way:
```
haskell/aura [ ./dist/build/aura/aura -Aabcdefghijklmnopqrstuvwxyz                                          ]  1 (8:03 PM)
Success (([AURInstall,DelMDeps,PacmanArg "dbpath" "cdefghijklmnopqrstuvwxyz"],[]),[])
aura >>= You have to use `sudo` for that.
```

fosskers:
No, that's evil. I wonder that's a bug in the library?

theotherjimmy:
I'll take a look and submit an issue if I need to.

theotherjimmy:
I just checked, and that is the same behavoir as pacman. eg.
```
haskell/aura [ pacman -bcdef                                          ]  255 (8:14 PM)
error: failed to initialize alpm library
(could not find or read directory: cdef)
```

fosskers:
I think I see what's happening here. If a short-op takes an arg, it tries to parse as much as possible out of that. Notice `-y` is a legal suboption for `-A`, but we see it parsed an argument to `-b`.

theotherjimmy:
I think it's done, but I would need to test all options. That would suck.

fosskers:
Haha very much so. I'm going to heavily review this as well.

theotherjimmy:
Just noticed, things like -Q don't work.

theotherjimmy:
This just might work.

theotherjimmy:
Looks good to me. It seems to do what you would expect: parse it's own commands and, if that fails, hand it off to pacman. Lets get this merged. @fosskers ?

PR #407, opened 2016-03-29
Suggested combined-search bash example change

gatlin:
At least on my versions of bash (4.3.42) and aura (1.3.4), the suggested combined search function won't execute the AUR search if the main repository search doesn't yield anything. This change ensures both run and display relevant results, if any.

### DIFF
--- a/doc/aura.8
+++ b/doc/aura.8
@@ -439,7 +439,7 @@ remove make deps, as well as show PKGBUILD diffs before building.
 use the following shell functions:
 .RS 4
 Bash => function search() {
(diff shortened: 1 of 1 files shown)


fosskers:
You need to update to `1.3.5`, as `1.3.4` is fundamentally broken due to the Arch folks not reregistering `aur4.archlinux.org` during an SSL certificate renewal.

bb010g:
You'd really want to just use a semicolon there instead to always make the AUR search run.

gatlin:
Even better!

PR #413, opened 2016-04-19
[Aura 1.4] Multiple fixes, updates from master

theotherjimmy:
pulled in the haskell-aur updates and fixed some of my bugs introduced by missuse of shelly. You should be able to build and install.


fosskers:
Travis seems to be puking?

theotherjimmy:
It's the 01-01-16 resolver. It's missing some deps.

fosskers:
You can remove that resolver from the travis config, then.

theotherjimmy:
Should be stable and look like the old aura in output now.

fosskers:
Minor bug with non-verbose output. The green colour from the `...` after the name of the package being currently built spills over onto the `makepkg` output. Can you confirm?

theotherjimmy:
yeah, that happens.

fosskers:
Should be auto-fix once we change how colouring works via `extensible-effects`. LGTM

PR #425, opened 2016-10-10
Create cache dir before trying to write to it

### LINKED ISSUE #409: `/var/cache/aura/pkgbuilds` does not exist after building aura, causing errors
First, I searched through the issues and couldn't find an instance of the operative keywords of the error in the issue tracker so please forgive me if this has been reported.

I pulled commit f06cf8e9a84a0ce8bf6f8505e151b6c46ca3edc4, and inside the directory ran

```
$> cabal sandbox init
$> cabal install --only-dependencies
$> cabal configure && cabal build
```

This built aura and I symlinked it to somewhere on my path. Worked great for searching and other querying. However, whenever I tried installing anything, I would get something similar to this:

```
$> sudo aura -A netatalk
[...]
aura: /var/cache/aura/pkgbuilds/netatalk.pb: openFile: does not exist (No such file or directory)
```

I fixed this by (as root) creating /var/cache/aura/pkgbuilds. I can look at the code and perhaps submit a pull request but I wanted to report this and see if anyone else had this issue. This is on a fresh Arch installation from today. I set up vim, git, and then pulled from master. So this is _not_ a complaint that an intermediate commit between releases had a bug, just that some kind of regression seems to be present in at least one case.

Or I'm a huge dumbass and I'll know why this is embarrassing soon enough. Either way, data!

theotherjimmy:
should resolve #409

### DIFF
--- a/aura/src/Aura/Pkgbuild/Records.hs
+++ b/aura/src/Aura/Pkgbuild/Records.hs
@@ -23,8 +23,8 @@ along with Aura.  If not, see <http://www.gnu.org/licenses/>.
 
 module Aura.Pkgbuild.Records where
 
-import System.Directory (doesFileExist)
-import System.FilePath  ((</>))
+import System.Directory (doesFileExist, createDirectoryIfMissing)
+import System.FilePath  ((</>), dropFileName)
 import Data.Foldable    (traverse_)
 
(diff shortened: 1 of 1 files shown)


theotherjimmy:
@gatlin does this work for you?

PR #431, opened 2017-02-06
fix curly brace extrapolations in dependencies

### LINKED ISSUE #430: Problem parsing dependencies, upgrade fails
Quite often, when I try to update AUR packages with aura, I get this type of error:

```
....
aura >>= Determining dependencies...
aura >>= Dependency checking failed for these reasons:
The dependency `perl-gnome2-wnckperl-gtk2-imageviewperl-gtk2-uniqueperl-x11-protocol` could not be found. You may need to search for a package to satisfy it.
```

If I update with any other AUR manager, it works out just fine. In the example above, for some reason `aura` concatenated all these dependencies into one and it failed to find them. I've noticed this about every second or third time I try to upgrade using `aura`.

Anyone know what's wrong? Thanks!

dpatti:
Fixes #430.

There are a few inconsistencies in the way bash strings are parsed that
make aura unable to correctly compile some AUR packages. A simple
example is the input:

    depends=(perl-{json,datetime})

is parsed as:

    ["perl-jsonperl-datetime"]

instead of:

    ["perl-json", "perl-datetime"]

This change puts in a few tweaks to fix this bug with extrapolation and
makes the bash parsing more consistent.

The first issue is that `DoubleQ` and `NoQuote` are `BashString`
constructors that contain a list of strings; however, it's not clear
what that list represents. The `unQuoted` parser treats the list of
strings as individual words, as `{foo,bar}` will be parsed as

    NoQuote [Right "foo", Right "bar"]

However, `doubleQuoted` parses expansions in the same way but later
concatenates them. So `"dep-${vers}"` will be parsed as

    DoubleQ [Right "dep-", Left (BashExpansion ...)]

Note that in the `NoQuote` example, you would expect a single string
with two words separated by a space, but in the `DoubleQ` example, you
would expect a single string with no spaces.

In a4bb0484b1, the change was made so that the list inside a `DoubleQ`
was concatenated as a singleton array, and I think this behavior makes
the most sense, since you can't represent variable interpolation if the
list represents individual words instead of fragments of a string. And
so, we must do the same when putting together `NoQuote`, too. This
effectively means that `{foo,bar}` will now produce `["foobar"]`, but we
can adjust the extrapolation behavior to accommodate for that.

There is also a strange behavior in `replaceStr`, where `NoQuote`
strings are turned into `DoubleQ` strings. As far as I can tell, this is
a mistake introduced in 95383612c9.

The last hurdle we must tackle is what `unQuoted` returns. In the case
of `{foo,bar}`, given our new understanding of what `NoQuote [...]`
means, we should be returning

    [NoQuote [Right "foo"], NoQuote [Right "bar"]]

We do this by pushing the `NoQuote` construction into the individual
parsers. This does however bring up an interesting point: the assignment
`depends=dep-${ver}` (without double quotes) will parse as an array of
two items. This was actually always the case given the way
`fromBashString` worked, so we're not introducing broken behavior. There
is more work that needs to be done on `unQuoted`, and perhaps some tests
would be extremely impactful.

### DIFF
--- a/aura/src/Bash/Base.hs
+++ b/aura/src/Bash/Base.hs
@@ -86,7 +86,7 @@ getVar ns s = case M.lookup s ns of
 fromBashString :: BashString -> [String]
 fromBashString (SingleQ s) = [s]
 fromBashString (DoubleQ l) = [fold $ rights l]
-fromBashString (NoQuote l) = rights l
+fromBashString (NoQuote l) = [fold $ rights l]
 fromBashString (Backtic c) = ["`" <> unwords (fromCommand c) <> "`"]
 
 fromCommand :: Field -> [String]
--- a/aura/src/Bash/Parser.hs
+++ b/aura/src/Bash/Parser.hs
@@ -128,9 +128,10 @@ expansion = char '$' *> choice [ BashExpansion <$> base <*> indexer <* char '}'
 
 -- | Replaces ${...}. Strings can be extrapolated!
 unQuoted :: Parser [BashString]
-unQuoted = fmap NoQuote <$> many1 ( choice [ try $ (: []) . Left <$> expansion
-                                           , fmap Right <$> extrapolated []
-                                           ])
+unQuoted = fold <$> many1 ( choice [ try $ (: []) . singleton . Left <$> expansion
+                                   , fmap (singleton . Right) <$> extrapolated []
+                                   ])
+  where singleton = NoQuote . (: [])
 
 -- | Bash strings are extrapolated when they contain a brace pair
 -- with two or more substrings separated by commas within them.
--- a/aura/src/Bash/Simplify.hs
+++ b/aura/src/Bash/Simplify.hs
@@ -77,7 +77,7 @@ replace' somethingElse = pure somethingElse
 replaceStr :: BashString -> State Namespace BashString
 replaceStr s@(SingleQ _) = return s
 replaceStr (DoubleQ s)   = DoubleQ . fmap Right <$> expandStr s
-replaceStr (NoQuote s)   = DoubleQ . fmap Right <$> expandStr s
+replaceStr (NoQuote s)   = NoQuote . fmap Right <$> expandStr s
 replaceStr (Backtic f)   = Backtic <$> replace' f
 
 expandStr :: [Either BashExpansion String] -> State Namespace [String]


dpatti:
I also _just_ realized that there is an effort in 1.4 to switch to `.SRCINFO` instead of `PKGBUILD`. I would be more than happy to help continue that work if someone would be interested in reviewing it.

fosskers:
Always interested in reviewing. The `1.4` branch work has stalled as of late, strictly because of time constraints.

fosskers:
Hindsight is 20/20. These days I'm fairly religious about writing tests for parsers/codecs, but the younger me (who wrote aura) was not so wise. That the Bash parser has no test suite is a continued source of annoyance. However, there's no point in writing one now, given that we're phasing the parser out entirely.

Is this PR a WIP? If so, prefixing the title with `[WIP]` would be helpful.

dpatti:
I was intending for this branch to be merged as-is. It at least fixes the common case of brace expansion without some sort of hack, and none of the small sample of tests that I was using locally broke.

fosskers:
Hm, this doesn't seem to work for the `streamlink-git` case:

```console
colin@yumi ~/c/h/aura> sudo .stack-work/install/x86_64-linux/nightly-2017-01-17/8.0.1/bin/aura -A streamlink-git
aura >>= Determining dependencies...
aura >>= Dependency checking failed for these reasons:
The dependency `python-` could not be found. You may need to search for a package to satisfy it.
```

dpatti:
You're right, but that's not any more broken than the current version. In order to fix that, we'd need to refactor a considerable amount and/or introduce some sort of `BashString = ... | MixedQuotes [BashString]` variant. All kinds of `"foo"bar` strings where adjacency means concatenation do not work.

dpatti:
(also, you might find `stack exec -- sudo aura -A ...` helpful)

fosskers:
It is helpful, thanks. I've been digging through my `.stack-work` manually up until now.

theotherjimmy:
Yeah, sorry about that refactor goof up. Good work finding it @dpatti !

fosskers:
```console
colin@yumi ~/c/h/aura> stack exec -- aura -Ai shutter
Repository  : aur
Name        : shutter
Version     : 0.93.1-8
AUR Status  : Up to Date
Maintainer  : PhotonX
Project URL : http://shutter-project.org/
AUR URL     : https://aur.archlinux.org/packages/shutter
License     : GPL3
Depends On  : xdg-utils imagemagick procps librsvg gnome-perl gtk2-perl desktop-file-utils perl-gnome2-wnck perl-gtk2-imageview perl-gtk2-unique perl-x11-protocol perl-proc-simple perl-proc-processtable perl-net-dbus perl-sort-naturally perl-json perl-json-xs perl-xml-simple perl-www-mechanize perl-locale-gettext perl-file-which perl-file-basedir perl-file-copy-recursive perl-path-class perl-xml-simple
Build Deps  : None
Votes       : 12
Popularity  : 8.68
Description : a featureful screenshot tool (formerly gscrot)
```
Expansions appear to work.

fosskers:
Although that package itself fails to resolve recursive deps, giving:
```
colin@yumi ~/c/h/aura> stack exec -- sudo aura -Axa shutter
[sudo] password for colin: 
aura >>= Determining dependencies...
aura >>= Dependency checking failed for these reasons:
The dependency `libgnome-data` could not be found. You may need to search for a package to satisfy it.
```

fosskers:
```
The dependency `libgnome-data` could not be found. You may need to search for a package to satisfy it.
colin@yumi ~/c/h/aura> stack exec -- sudo aura -Ai libgnome-data
aura: Maybe.fromJust: Nothing
```
Usually this occurs when there are some clever virtual package workings happening deep somewhere.

fosskers:
Nope, [it's a legit package](https://aur.archlinux.org/packages/libgnome-data).

dpatti:
Interesting. I'll try and take a closer look. For what it's worth, this seems to work fine when I build from the branch, but I don't know if the state of my machine is influencing that.

```
~/d/a/aura ❯❯❯ stack exec -- sudo aura -Axa shutter
aura >>= Determining dependencies...
aura >>= AUR Packages:
shutter

aura >>= Continue? [Y/n] 
aura >>= Building `shutter`...
```

fosskers:
It's `gnome-vfs` which is provided by `https://aur.archlinux.org/packages/gnome-vfs-nosmb/` but is not itself a package.

dpatti:
Interesting. It certainly _used_ to be. I have a copy in my pacman cache:

```
~/d/a/aura ❯❯❯ tar xf /var/cache/pacman/pkg/gnome-vfs-2.24.4-10-x86_64.pkg.tar.xz .PKGINFO -O
# Generated by makepkg 5.0.1
# using fakeroot version 1.21
# Wed Aug  3 14:08:12 UTC 2016
pkgname = gnome-vfs
pkgver = 2.24.4-10
pkgdesc = The GNOME Virtual File System
url = http://www.gnome.org
builddate = 1470233292
packager = Jan de Groot <jgc@archlinux.org>
size = 5793792
arch = x86_64
license = LGPL
depend = gconf
depend = bzip2
depend = avahi
depend = smbclient
depend = gnome-mime-data
depend = krb5
depend = gnutls
depend = libgcrypt
makedepend = pkgconfig
makedepend = intltool
makedepend = gtk-doc
makedepend = gnome-common
```

fosskers:
The shutter dep `libbonoboui` is failing to build for me.

fosskers:
Shutter doesn't build for me personally, but that's for a different reason that what this PR addresses. This PR does produce a confirmed fix for bash expansions, so in it goes.

PR #432, opened 2017-02-07
pull out translated field sets into individual translations

dpatti:
This takes three translation functions that return an array of
translated strings and pulls out each of the strings into their own
translation function. Then we can get the same result by using
`sequence`, since the signature looks like this:

    sequence :: (Traversable t, Monad m) => t (m a) -> m (t a)

or, in this instance, where `m` is `(-> Language)` and `t` is `[]`:

    sequence :: [Language -> String] -> (Language -> [String])

There are good reasons for doing this. For one, it is very hard to add a
field to the list. You have to figure out where to insert it once for
_each language_. This is not only tedious, but it is error prone, too.
For example, in `infoFields`, the spanish translation list was lacking
an entry for "Popularity", which meant that it would label the
popularity field as "Descripción" and probably drop the description
field completely. And then, after you figure out the index in which to
insert it, you must place the english label as a placeholder for each
translation, which makes it hard to see what needs translating. As part
of this change, fields that were placeholders were removed so that it is
easy to see what is missing.

The other reason for this is that some fields are shared, like many of
the fields between `infoFields` and `absInfoFields`. Because the strings
were duplicated, there were some translations available in one but not
the other. The deduplicated fields in this change include translations
from both sources. There were no conflicting translations to resolve.

### DIFF
--- a/aura/src/Aura/Languages.hs
+++ b/aura/src/Aura/Languages.hs
@@ -152,6 +152,279 @@ langFromEnv = \case
     _    -> English
 
 ----------------------
+-- Field labels
+----------------------
+
+package :: Language -> String
+package = \case
+    Japanese   -> "パッケージ"
+    Polish     -> "Pakiet"
+    Croatian   -> "Paket"
+    Swedish    -> "Paket"
+    German     -> "Paket"
+    Spanish    -> "Paquete"
+    Portuguese -> "Pacote"
+    French     -> "Paquet"
+    Russian    -> "Пакет"
+    Italian    -> "Package"
+    Serbian    -> "Пакет"
+    Norwegian  -> "Pakke"
+    Indonesia  -> "Paket"
+    _          -> "Package"
+
+firstInstall :: Language -> String
+firstInstall = \case
+    Japanese   -> "初インストール"
+    Polish     -> "Pierwsza instalacja"
+    Croatian   -> "Prva instalacija"
+    Swedish    -> "Första installation"
+    German     -> "Erste Installation"
+    Spanish    -> "Primera instalación"
+    Portuguese -> "Primeira instalação"
+    French     -> "Première installation"
+    Russian    -> "Первая установка"
+    Italian    -> "Prima installazione"
+    Serbian    -> "Прва инсталација"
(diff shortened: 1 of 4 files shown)


dpatti:
The reason I wanted to do this was because I noticed that `aura -Ai gnome-vfs-nosmb` was not showing that `gnome-vfs-nosmb` provides `gnome-vfs`, which was confusing until I looked at the pkgbuild. But when I went to add a new field to the info output, I got very concerned about botching the translations. In the process of refactoring, I found a few errors, so I thought this was a good change to make.

Most of the code introduced here was generated automatically with a little massaging by hand. If you would like the naming scheme changed at all (e.g., `package` -> `packageField`), I would be happy to make the changes.

fosskers:
Thanks @dpatti , I'm going to go over this on the weekend.

aura/src/Aura/Languages.hs
```diff
@@ -39,31 +39,19 @@ along with Aura.  If not, see <http://www.gnu.org/licenses/>.
 
 -}
 
-module Aura.Languages where
+module Aura.Languages
+    ( module Aura.Languages
```
fosskers: If this is a way to reexport the two new modules you've added, it would be better to make that explicit a la:

```haskell
( module Aura.Languages.Base
, module Aura.Languages.Fields
...
```

dpatti: The problem is that we need to export `Language(..)` from `Base`, but if we specify that export, then all of the functions in `Aura.Languages` are not exported. So we either need to list each function or have the module export itself. The way I have it, `Fields` is not exported at all.

We could always go the route of moving all the translations to another module and keeping `Languages.hs` slim to avoid the self-export.

fosskers: [Indeed!](https://stackoverflow.com/questions/17241491/haskell-hide-specific-functions-in-module-exports)

I see the issue now. Your currently formulation is fairly clean then, given the circumstances.

fosskers:
I can confirm that this fixes the "bumped field" problem, as you mentioned with `--spanish`.

PR #434, opened 2017-02-09
allow use of language flags

dpatti:
Currently, when you use any language flag with a command, such as

    aura -Ai aura --german

you get the error

    aura >>= Angegebene Kommandozeilen-Flags sind widersprüchlich!

which is the german translation of "Conflicting flags given!". This is
because `executeOpts` in the main executable module is strict about
which flags it is given for a command. Before `executeOpts` is called,
though, all settings flags, such as `--quiet` are stripped out so that
they do not cause conflicts. It appears that language flags should also
be stripped out, and doing so prevents this conflicting error.


fosskers:
Excellent, these work again. I noticed that the Indonesian translator didn't actually add a flag for their use, so I'll add one now.

dpatti:
@fosskers It's also worth noting that you can close #343 as well (since I believe it was trying to fix the same problem)

PR #437, opened 2017-02-10
update 1.4 with master (including 1.3.7)

dpatti:
There have been some major diversions on both ends, so this brings all
the recent 1.3 work into the 1.4 branch. To summarize:

* The `aur` package was updated to 6.0.
* The `nightly-2016-10-01` is used as the stack resolver. Note that
  `megaparsec` and `basic-prelude` had some major changes in the
  interim, so they have been locked to their old `lts-5.13` versions for
  the time being.
* `wreq` has been replaced with `http-client`, and the functions that
  take `Manager` objects have been updated.
* `microlens` is no longer a dependency.

The most tedious part of the merge was dealing with all the import
conflicts. I tried for the most part to stay true to the style of the
current master in terms of ordering and spacing, but there are likely
still some differences.

---

I thought I would do this to get an understanding of how the codebase has changed in both directions since 1.4 was started. It was quite a rocky journey, but I think I have a better understanding of where things are now. I suppose there are a few options going forward:

1. Proceed with the `aura-1.4` branch, dealing with merge conflicts like this. The biggest downside is that there could be a huge barrier to working on 1.4 in the future if things diverge again.
2. Merge `aura-1.4` into `master` and optionally keep a `1.3.x` branch for further patch releases if necessary. This has the added benefit of putting most new code into `master` and only important changes can be applied to the `1.3.x` branch, released, and merged back into master immediately.
3. Abandon the `aura-1.4` branch, but use it as a guide, and split it into smaller prs (i.e., using the top-level checklist items) that get merged into `master` as they are done. This has the added benefit of creating a less confusing history, and as long as each branch builds and is tested, there is the option of not even having a 1.3 branch.

To me, the outsider, the third option sounds the most attractive, mostly because even this merge commit looks horrifying, when I actually (hopefully) did not refactor or add any new behavior. It's something to think about, I suppose.


fosskers:
Thank you for your effort, I'm sure this PR wasn't without pain. I'm going to mull-over our options. `1.4` adds some much-needed updates, though the whole project is dying for a modern rewrite from scratch anyway (this thought has been at the back of my mind for a few years). 

The potentially confusing spacing issues surrounding import statements is a reflections of Emacs's `haskell-mode` automatically formatting them like that. Something you've probably seen:

```haskell
 import qualified Foo as F
 import           Bar (bar)
```
I use this style now with all my Haskell libraries, but a few of Aura's older modules might not have had this done to them.

dpatti:
I've seen and used the `qualified` spacing, but one thing that was a little new to me was spacing between the module and symbols, like here: https://github.com/aurapm/aura/blob/aura-1.4/aura/src/aura.hs#L53. Sometimes there will be a sequence of a few imports that line up nicely, but I can't tell exactly what the pattern is.

As far as a rewrite goes, I would definitely be willing to help, but I suppose you have most of the big picture floating around in your head. However, the reason I've been trying to contribute (in addition to having some time on my hands) is that the current version of Aura doesn't handle some packages correctly, including some of the packages that I have installed on my machine. I would really love to have a version of Aura that deals with `SRCINFO`, and though I don't know exactly what that entails, I was hoping that it's not too far out of reach.

PR #442, opened 2017-02-16
Fixing `-B`

fosskers:
There were some oversights in how `-B` flags are handled. Critically, this PR removes `-r` and `-b` (for `-C`) as subflags, as there was some unresolvable conflicts with underlying pacman flags. Their long forms can be used instead, and the man page has been changed to reflect that.


fosskers:
@HaleTom these fixes will be included in `1.3.8`.

PR #446, opened 2017-03-05
add Chinese translation

kaizhang-shanxi:
Hi, I add Chinese translation in this PR, hope it helps:)

### DIFF
--- a/aura/src/Aura/Flags.hs
+++ b/aura/src/Aura/Flags.hs
@@ -124,6 +124,7 @@ data Flag = ABSInstall
           | SerbianOut
           | NorwegiOut
           | IndoOut
+          | ChineseOut
           | PacmanArg String String
             deriving (Eq, Ord, Show)
 
@@ -231,6 +232,7 @@ languageOptions = fmap simpleOption
                   , ( [], ["serbian", "српски"],       SerbianOut  )
                   , ( [], ["norwegian", "norsk"],      NorwegiOut  )
                   , ( [], ["indonesian"],              IndoOut     )
+                  , ( [], ["chinese", "中文"],         ChineseOut  )
                   ]
 
 -- `Hijacked` flags. They have original pacman functionality, but
@@ -272,7 +274,7 @@ settingsFlags = [ Unsuppress, NoConfirm, HotEdit, DiffPkgbuilds, Debug, Devel
 languageFlags :: [Flag]
 languageFlags = [ JapOut, PolishOut, CroatianOut, SwedishOut, GermanOut
                 , SpanishOut, PortuOut, FrenchOut, RussianOut, ItalianOut
-                , SerbianOut, NorwegiOut, IndoOut ]
+                , SerbianOut, NorwegiOut, IndoOut, ChineseOut ]
 
 -- Flags like `AURIgnore` and `BuildPath` have args, and thus can't be included
 -- in the `settingsFlags` list.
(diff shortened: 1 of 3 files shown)


fosskers:
Amazing, I've been hoping for these! Thank you very much.

fosskers:
I see you've provided simplified characters. Should we use `--mandarin` and `--简体字` instead? I know Cantonese and Taiwanese dialects use traditional characters.

kaizhang-shanxi:
You are right, some Cantonese(mainly in Hong Kong) and [Taiwanese(a variant of Hokkien)](https://en.wikipedia.org/wiki/Taiwanese_Hokkien) use traditionnal characters. However, there is one point to clarify: mandarin refer to the spoken language, not the characters. Actually, [Chinese characters divide into simplified form and traditional form](https://en.wikipedia.org/wiki/Chinese_characters).

And I note that aura only takes the first 2 characters of the locale. So my advice is:

+ If aura will keep this logic, I think `--chinese` and `--中文` is proper, because most of the people who write Chinese characters use the simplified form
+ If aura will parse the whole locale string in the future, I think `--zh-cn`(`zh_CN` in /etc/locale.gen) and `--简体中文`) is proper. Because `简体字` doesn't contain the meaning of **Chinese**

Thank you for your appreciation :)

fosskers:
I agree with you that the vast majority of people who speak Chinese languages speak Mandarin. Certainly in English we mean "Mandarin" when we say "Chinese", and the average person probably isn't aware of the difference, or that Chinese is actually a large language family. 

For Aura (and because I love languages), I like to be as literal as possible. I have learned a little bit of 廣東話, but no 普通话. What do you think of:

- Your translation: `--mandarin` + `--普通话`
- Possible future translation: `--cantonese` + `--廣東話`
- Possible future translation: `--taiwanese` + `--臺灣話`

I think this gives us the most potential for the future. Do you think Chinese users would expect to `--chinese` or `--mandarin`? Would they prefer one over the other? Or would Chinese users mainly use the Kanji versions?

kaizhang-shanxi:
Yes, [Chinese is a large language family](https://en.wikipedia.org/wiki/Chinese_language). Besides Yue(including Cantonese) and Min(including Taiwanese), Chinese language contains Wu, Gan, Xiang, and Hakka, etc. However, although a variety of **spoken** language exist, there are only two form of **written** Chinese, i.e., the simplified one and the traditional one. And I think we need written language here.

It's no doubt that Chinese users do prefer `--chinese`. I known Mandarin is widely used in western world, but for Chinese people, we will feel confused about it. Because Mandarin, I mean this word, is rarely seen in Chinese textbooks, TVs or official documents.

As for Kanji, [it's a Japanese version of Chinese characters](https://en.wikipedia.org/wiki/Kanji), and mainly used by Japanese, not Chinese.

fosskers:
> And I think we need written language here.

You're right, it's been my experience that Cantonese speakers are literate in simplified written Chinese, even if they can't speak it very well. It surprised me the first time I saw a Cantonese TV show whose subtitles were in simplified form, and using the mandarin forms for certain words (i.e. `是` instead of the Cantonese `係`). In that case, I propose we use all three of the following flags for your translation:

- `--chinese`
- `--simplified-chinese`
- `--中文`

> As for Kanji...

I said Kanji because I forgot what the chinese word for them is :P We write it as `漢字` as in "the characters of the Han people", even though they were first borrows during the Tang period, haha.

Jiehong:
Ah, you should be cautious with namings…
For example, in Taiwan most people do speak mandarin, and they write it with traditional characters that are different from the traditional characters used in Hong-Kong for example.
Yet, some people (in the South of the island mainly), also speak Taiwanese, which is a dialect is almost never written. It's actually ambiguous on how to write it, as mostly traditional Chinese characters are used either for a close sound or close meaning, and some don't exist elsewhere.

It's safe to say that you only want to differentiate the writing form and its relation to the place, as characters may differ.

I'd offer:

`--chinese-prc/--简体中文` : for simplified Chinese used in China
`--chinese-hk/--繁體香港` : for traditional Chinese used in HK
`--chinese-tw/--繁體台灣` : for traditional Chinese used in Taiwan

It's was is used on my phone keyboard for example.

kaizhang-shanxi:
`--chinese`, `--simplified-chinese` and `--中文` look good to me :P. @fosskers 

As you say, most people in Taiwan Province of China speak Mandarin, and they write it with traditional characters. And the traditional Chinese in Taiwan Province and that in Hong Kong are not identical. However, most traditional characters are the same, only a few differ. What's more, I have never seen translations such as chinese-prc, chinese-hk, and chinese-tw before. For example,

+ [Here](https://en.wikipedia.org/wiki/List_of_Wikipedias#List) is the language list of Wikipedia, the Chinese translations are Chinese and Classical Chinese
+ The left panel of [Arch Linux Wiki] list its supported languages, the Chinese translations are 中文（简体） and 中文（繁体）
+ [Here](http://www.un.org/en/sections/about-un/official-languages/) is official languages of UN, the translation is Chinese

So I don't think chinese-prc, chinese-hk, and chinese-tw are OK for the Chinese translation. @Jiehong

fosskers:
@kaizhang-shanxi thanks for this!

PR #454, opened 2017-07-05
Added logo in SVG.

Anton-Latukha:


fosskers:
Awesome, thank you! We should use this where ever we can.

bb010g:
Would it be possible to keep the yellow / gray / dark gray color scheme currently used? The blue eye feels jarring against the yellow, and the old gray logo colors fit better with a gray eye.

Anton-Latukha:
A coloring of the ready picture is very simple in Inkscape.
Haskell logo is grouped, ungroup and change filling colors.
I leave further coloring up to project.

bb010g:
We're going to see about redoing this.

PR #477, opened 2018-02-05
Link between no-orphans and working-with-orphans

ahstro:
Add link between "No orphans" and "Working with orphan packages", so one is not left wondering how Aura "helps keep track of and remove [them]".


fosskers:
Beauty, thank you.

PR #482, opened 2018-06-22
Update Portuguese translation

wamaral:
Included some missing translations, and fixed some that were not very precise

Since it's only a few changes, I didn't include my name in the list. Feel free to do so if you'd like, though


fosskers:
Thank you! Actually, Aura is undergoing a massive overhaul. Can you target this to the `aura-1.5` branch to make sure everything is correct?

wamaral:
Yeah, sure! I'll fix it real quick

wamaral:
Done :)

fosskers:
Awesome :sunglasses: I'll merge this on Monday (24 hours from now or so).

fosskers:
Thanks again for these.

PR #486, opened 2018-07-04
Do not exit with an error if there is nothing to upgrade

### LINKED ISSUE #315: Exit with success for `aura -Ayu` if no packages are being upgraded/updated
First of, thanks for this great tool, I was looking for something that can automate package installs since _yaourt_ does not work as root at the moment, so I stumbled upon this amazing thing that does even more than I was actually looking for (having checkpoints with `-B` is simply genius :smiley:)

---

At the moment aura returns 1, though there is actually no error, its just that everything is up to date. That complicates scripting.

### LINKED ISSUE #484: Exit with success if no packages need to be upgraded
In contrast to Pacman, Aura exits with a non-zero error code when there are no packages to be upgraded. 

```shell
$ sudo pacman -Su      
:: Starting full system upgrade...
 there is nothing to do
$ echo $?                           
0
$ sudo aura -Au  
aura >>= Fetching package information...
aura >>= Comparing package versions...
aura >>= No AUR package upgrades necessary.
aura >>= No valid packages specified.
$ echo $?
1
```

There is no error here so the error code should be 0 (success).

koterpillar:
Fixes #315.
Fixes #484.

### DIFF
--- a/aura/lib/Aura/Commands/A.hs
+++ b/aura/lib/Aura/Commands/A.hs
@@ -87,9 +87,10 @@ upgradeAURPkgs pkgs = do
          devel <- develPkgCheck
          send . notify ss $ upgradeAURPkgs_2 lang
          if | null toUpgrade && null devel -> send . warn ss $ upgradeAURPkgs_3 lang
-            | otherwise -> reportPkgsToUpgrade toUpgrade (toList devel)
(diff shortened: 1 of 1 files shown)


fosskers:
Thank you! You beat me to the punch.

PR #488, opened 2018-07-05
Fix language names

koterpillar:
`languageNames` was updated sloppily at some point in the past - it has less language names for one of the languages. I've added a test so that doesn't repeat in the future and added the missing value.

```shell
$ LANG=ru stack exec aura -- --version
                       
 __ _ _  _ _ _ __ _    Pacman v5.1.0 - libalpm v11.0.0
/ _` | || | '_/ _` |   Copyright (C) 2006-2018 Pacman Development Team
\__,_|\_,_|_| \__,_|   Copyright (C) 2002-2006 Judd Vinet
AURA Version 1.4.0     
 by Colin Woodbury     This program may be freely redistributed under
                       the terms of the GNU General Public License.
Переводчики Aura:      
 Chris "Kwpolska" Warrick (Польский)
 Denis Kasak / "stranac" (Хорватский)
 Fredrik Haikarainen / Daniel Beecham (Шведский)
 Lukas Niederbremer / Jonas Platte (Немецкий)
 Alejandro Gómez / Sergio Conde (Испанский)
 Henry "Ingvij" Kupty / Thiago "thiagowfx" Perrotta (Португальский)
 Ma Jiehong / Fabien Dubosson (Русский) (*)
 Kyrylo Silin (Итальянский)
 Bob Valantin (Сербский)
 Filip Brcic (Норвежский)
 "chinatsun" (Индонезийский)
 "pak tua Greg" (Chinese)
aura: Map.!: given key is not an element in the map
CallStack (from HasCallStack):
  error, called at libraries/containers/Data/Map/Base.hs:489:16 in containers-0.5.7.1:Data.Map.Base
```

Also, I've started updating some of the translations - if you'd like, I'll do that as a separate PR with a complete update (of what I speak).

### DIFF
--- a/aura/lib/Aura/Languages.hs
+++ b/aura/lib/Aura/Languages.hs
@@ -85,12 +85,12 @@ languageNames = Map.fromList . zip [ Japanese, Polish, Croatian, Swedish, German
     Spanish    -> [ "Japanese", "Polaco", "Croata", "Sueco", "Alemán", "Español", "Portugués", "Francés", "Ruso", "Italiano", "Serbio", "Noruego", "Indonesio", "Chinese" ]
     Portuguese -> [ "Japonês", "Polonês", "Croata", "Sueco", "Alemão", "Espanhol", "Português", "Francês", "Russo", "Italiano", "Sérvio", "Norueguês", "Indonésio", "Chinês" ]
     French     -> [ "Japanese", "Polonais", "Croate", "Suédois", "Allemand", "Espagnol", "Portugais", "Français", "Russe", "Italien", "Serbe", "Norvégien", "Indonesian", "Chinese" ]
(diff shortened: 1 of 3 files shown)


koterpillar:
Also, for proper localization of messages with numbers in them, some languages will need complex [pluralization rules](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals) (TL;DR: one "singular" and one "plural" form are not enough, there are languages with anywhere up to 6).

Do you have an opinion on bringing in a dependency on something like https://www.yesodweb.com/book/internationalization? The alternative would be to define the pluralization rules in `Languages.hs` as well, or some ugly messages.

fosskers:
Thank you! 

I'm on the fence about the translations. Having all the strings directly in the executable definitely bloats it, but it's also nice to get compile-time guarantees about the messages. Also, all translations are right next to each other, which could be handy for multilingual translators.

fosskers:
Hopefully I didn't merge this prematurely.

fosskers:
As for getting access to the `RenderMessage` typeclass, that seems to be from `shakespeare`, which has a big host of dependencies that I'm not too keen to pull into Aura.

PR #491, opened 2018-07-24
README.md: fix typo "privileges"

AladW:
https://www.merriam-webster.com/dictionary/privilege

Note that the README also mentions
>Built AUR package files are moved to the package cache. This allows for them to be easily downgraded when problems arise. Other top AUR helper programs do not do this

which is false; at least aurutils moves everything to a local repo in `/var/cache/pacman/<something>`. I left this for a different commit, though.


fosskers:
What is meant by the "other helpers" comment is referring to the ability to easily perform downgrades, but I agree that the wording is ambiguous.

PR #508, opened 2018-08-27
AUR Security Sweep

fosskers:
This PR adds a side-executable that scans AUR packages for security flaws (or just bad bash).

Out of ~50k packages currently on the AUR, it flags 1,328 as containing bad bash.

### DIFF
--- a/aur-security/Main.hs
+++ b/aur-security/Main.hs
@@ -1,0 +1,55 @@
+{-# LANGUAGE LambdaCase, NamedFieldPuns, DuplicateRecordFields, TupleSections #-}
+
+module Main ( main ) where
+
+import           Aura.Packages.AUR (aurLookup)
+import           Aura.Pkgbuild.Security (parsedPB, bannedTerms)
+import           Aura.Types
+import           BasePrelude
+import           Control.Compactable (fmapEither)
+import           Control.Concurrent.STM (atomically)
+import           Control.Concurrent.STM.TQueue (flushTQueue)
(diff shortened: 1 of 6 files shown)


maximbaz:
@fosskers this is awesome idea, have you considered extracting this in a standalone project? For example, it would have been so cool to install this executable alone and make it part of `aurutils` pipeline. I'm thinking about a tool that would accept path to the folder with PKGBUILD as an argument, so that all the downloading magic remains in aura.

/cc @AladW

fosskers:
Now that `aura` is a library too, it's fairly easy to depend on it and use the `Aura.Pkgbuild.Security` module, as `aur-security` is. If you're not a Haskell dev, I could produce such a tool fairly easily for you, or just expand on `aur-security`, giving it different CLI flags for different operation modes, say.

maximbaz:
Great! Yeah, I'm not good at Haskell, so if you can help with the tool, it would be awesome 😉 

I'm thinking about the following: how to keep the tool up-to-date (if you change security module, it needs to be rebuilt), and how to make it easy to install for people (e.g. publish on AUR). I initially suggested to split the code and extract all security-related stuff in a separate repo, but honestly if you feel it should all remain together, I don't mind having `aura` installed even if I use it only for security-related features in combination with another AUR helper.

For the tool itself, I would say here's the basic need:

1. Accept path to folder with PKGBUILD via CLI argument, e.g. `$ aur-security /tmp/mypackage/`
2. Return 0 exit code if no security issues are found and non-zero exit code if there are issues
3. Print the list of issues on stdout

PR #512, opened 2018-10-07
Add Esperanto translations

Ghosy:
Please let me know if there is anything that needs to be done that wasn't included in the localization guide.

### DIFF
--- a/aura/lib/Aura/Languages.hs
+++ b/aura/lib/Aura/Languages.hs
@@ -27,6 +27,7 @@ Filip Brcic                                   | Serbian
 "pak tua Greg"                                | Indonesia
 Kai Zhang                                     | Chinese
 Onoue Takuro                                  | Japanese
+Zachary "Ghosy" Matthews                      | Esperanto
 -}
 
 module Aura.Languages where
@@ -60,6 +61,7 @@ translators = Map.fromList
     , (Indonesia,  "\"pak tua Greg\"")
     , (Chinese,    "Kai Zhang")
     , (Japanese,   "Onoue Takuro")
+    , (Esperanto,  "Zachary Matthews")
     ]
 
 -- These need updating! Or removing...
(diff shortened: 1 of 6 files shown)


aura/lib/Aura/Types.hs
```diff
@@ -218,6 +218,7 @@ data Language = English
               | Norwegian
               | Indonesia
               | Chinese
+	      | Esperanto
```
fosskers: Tiny little alignment issue here.

aura/lib/Aura/Languages.hs
```diff
@@ -79,6 +81,7 @@ languageNames = Map.fromList . zip [ Japanese, Polish, Croatian, Swedish, German
     Norwegian  -> [ "Japanese", "Polsk", "Kroatisk", "Svensk", "Tysk", "Spansk", "Portugisisk", "Fransk", "Russisk", "Italiensk", "Serbisk", "Norsk", "Indonesian", "Chinese" ]
     Indonesia  -> [ "Japanese", "Polandia", "Kroasia", "Swedia", "Jerman", "Spanyol", "Portugis", "Prancis", "Rusia", "Italia", "Serbia", "Norwegia", "Indonesian", "Chinese" ]
     Chinese    -> [ "日语", "波兰语", "克罗地亚语", "瑞典语", "德语", "西班牙语", "葡萄牙语", "法语", "俄语", "意大利语", "塞尔维亚语", "挪威语", "印度尼西亚语", "中文" ]
+    Esperanto  -> [ "La japana", "La pola", "La kroata", "La sevda", "La germana", "La hispana", "La portugala", "La franca", "La rusa", "La itala", "La serba", "La norvega", "La indonezia", "La ĉina"]
```
fosskers: In Esperanto, is there a way to refer to language names without using a definite article? If not, then it's fine the way you have it here.

Ghosy: The article is needed in Esperanto because it provides the context that you are referring to the shortened form for a language(La angla = La angla lingvo). Without la you lose the context to show you are referring to the language.

aura/lib/Aura/Types.hs
Ghosy: Fixed by 4a5f498.

fosskers:
Thank you!

PR #524, opened 2019-01-19
Uses fused-effects instead of freer-simple

jkachmar:
What it says on the tin above, this is pretty much a rote translation from `freer-simple` to `fused-effects`.

I tried to riff off of robrix/fused-effects#82 a bit for the `sendM` thing, but I'm still having some trouble getting it to work with `MonadIO` rather than plain ol' `IO`.

I don't have Arch installed and haven't gotten to setting up a Docker image to let me thoroughly test and profile the executable in.


fosskers:
[These are some nice numbers.](https://github.com/patrickt/effects-benchmarks) I will get this merged.

jkachmar:
This is a bit out of date; when I get home tonight I’ll see about targeting the latest `fused-effects` and cleaning it up.

fosskers:
Thanks man :+1:

fosskers:
Note on the `package.yaml` conflict: I've dropped usage of those files and gone back to `.cabal` files for each of the subprojects here.

jkachmar:
I've added `fused-effects-0.4.0.0` as an extra dependency, since it's got all the updated goodies, rebased off of develop, and chased down all the type errors.

As a small stylistic change I've used explicit imports for all the effects, even though `Control.Effect` re-exports a lot, just so it's easier to track what comes from which module.

fosskers:
Thank you!!

jkachmar:
I’ll have to check tomorrow, but I think the `pPrintNoColor` workaround I added isn’t necessary anymore (given that `Lift` has a `MonadUnliftIO` instance).

If you get around to merging this beforehand I can open a very small PR afterwards, otherwise I’ll just update it here :)

fosskers:
I'll need another day or two to get around to finishing #530 .

fosskers:
... I merged #530 and broke everything here, but the looks of it! :sweat_smile:

jkachmar:
Okay, that was waaaaaaaaay more annoying that I thought it was going to be.

It compiles locally, but there's always the chance that I messed up something when merging and translating the type sigs to be `fused-effects`-friendly, so one more review would probably be helpful.

fosskers:
I'll be double-checking all of it :) Thank you, my friend.

fosskers:
Just merged another big thing (sorry). I'll take this over, fix up the conflicts, and then get some perf numbers.

fosskers:
Aaaand done!

jkachmar:
🎉 thanks for the merge!

Did you get a chance to take perf numbers, and if so how do they look?

fosskers:
I didn't bother generating a flame graph, but I did my usual timing of dependency resolution of `ros-lunar-desktop`, which one of the biggest dep graphs on the entire AUR. Resolution speed didn't slow down at all with this PR (and everything still works).

fosskers:
I'll be doing a release soon, so that this work is included in an official release.

PR #544, opened 2019-08-31
Fix esperanto name

### LINKED ISSUE #542: error Map bad index access when doing --version on current master
~~~~
> aura --version
                       
 __ _ _  _ _ _ __ _    Pacman v5.1.3 - libalpm v11.0.3
/ _` | || | '_/ _` |   Copyright (C) 2006-2018 Pacman Development Team
\__,_|\_,_|_| \__,_|   Copyright (C) 2002-2006 Judd Vinet
AURA Version 2.0.1     
 by Colin Woodbury     This program may be freely redistributed under
                       the terms of the GNU General Public License.
Aura Translators:      
aura: Map.!: given key is not an element in the map
CallStack (from HasCallStack):
  error, called at libraries/containers/Data/Map/Internal.hs:610:17 in containers-0.6.0.1:Data.Map.Internal
~~~~

koterpillar:
Same as #488 before. Turns out the test there was incomplete, I've updated it this time.

Fixes #542.


stack.yaml.lock
```diff
@@ -0,0 +1,75 @@
+# This file was autogenerated by Stack.
```
fosskers: This file shouldn't be committed. You can do a `git reset --soft` to open the commit back up, remove the change, and then force-push.

fosskers: Ah nevermind, I see you specifically committed this in a single commit. Hm, then you'd have to do some rebasing.

koterpillar: Stack recommends [committing the lock file](https://docs.haskellstack.org/en/stable/lock_files/) for reproducibility. Is there a reason it shouldn't be committed in this project?

fosskers: Oh, that's new information for me. In that case, let's keep it.

koterpillar:
Thanks for approving! Unfortunately, I can't merge anything here myself :-)

fosskers:
I'll get to it.

fosskers:
Ĉu mi diris, ke ankaŭ mi lernas la Esperanton? ;)

PR #551, opened 2019-11-05
Updated Spanish language

PandaFoss:
I have updated most of the lines to Spanish. On the end of the file there are some that are only in English, should they also be translated?


fosskers:
Thank you for this! Yes, please add translations whereever they're missing.

PandaFoss:
I think I have finished the Spanish translation of all the lines.
I have an other question. In the `Languages.hs` file, on line 1359 I think it is necessary to add `\case` to the end of the line, in order to add the translations, am I correct?

fosskers:
Which function name, specifically? `provides_1`? Then yes, You'd need the `\case` before pattern matching as is done in the other functions.

PandaFoss:
I add it?

fosskers:
Yes please :)

PandaFoss:
Hmm, there was a problem on the last checkup but I'm not sure...

fosskers:
It's okay, I'll fix it and merge it manually. Thank you!

fosskers:
Thanks again!

PR #598, opened 2020-05-15
Persist VCS Sources

fosskers:


aura/lib/Aura/Install.hs
```diff
@@ -185,11 +185,17 @@ buildAndInstall bss = do
...
+          -- to look it up in the package cache. If we find a match, we don't
+          -- need to build it.
+          g :: Buildable -> Either Buildable PackagePath
           g b = case bToSP b `M.lookup` cache of
             Just pp | not (switch ss ForceBuilding) -> Right pp
             _                                       -> Left b
+
       built <- traverse buildPackages $ NEL.nonEmpty ps
-      traverse_ installPkgFiles $ fmap (<> cached) built >>= NEL.nonEmpty
+      traverse_ installPkgFiles $ (built <> Just cached) >>= NEL.nonEmpty
```
fosskers, line 198: There was a subtle bug here.

PR #601, opened 2020-05-21
Nixpkgs dev. support & static-haskell-nix builds

bb010g:
The former commit is just enough to get `nix-shell` or lorri happy, and it works fine.

The latter commit gets us static builds with [nh2/static-haskell-nix](https://github.com/nh2/static-haskell-nix), and it's currently blocked on strange segfaults from GCC+Musl C binaries (nh2/static-haskell-nix#73).


bb010g:
Okay, figured out the segfault problem (n.b.: it's not really figured out, but it's worked around enough). Will be pushing that in a bit.

bb010g:
Ready to merge.

fosskers:
Can you elaborate on what each file is, and how this is used?

I'm particularly curious about that giant `aura-stack2nix-output.nix`. Does that need to be committed?

bb010g:
`/nix/aura-stack2nix-output.nix` is handy to commit because generation can take a bit. It's basically the Stackage snapshot we're using, plus any extra Hackage packages at a specified timestamp, plus our packages. It shouldn't bloat statistics here thanks to the linguist generated attribute in `.gitattributes`, though.

`/nix/default.nix` is common definitions (set up to be nicely overrideable at the command line). `/nix/nixpkgs-ghc-2234-cross-detect.patch` is there because I'm not sure if we won't need it to fix a cross-related bug that came up before the next (unreleased) GHC release, and until we're past that version I'd prefer to keep it at hand. `/nix/sources.json` & `/nix/sources.nix` handle dependency version pinning via [niv](https://github.com/nmattia/niv). `/nix/stack.nix` works with Stack's Nix integration. `/nix/stack2nix-per-ghc-base-packages.patch` patches [stack2nix](https://github.com/input-output-hk/stack2nix) to work with base packages it didn't know about by generating them off of the data at https://gitlab.haskell.org/bgamari/ghc-utils. `/nix/static-haskell-nix-fixed_stack2nix-override.patch` allows us some stack2nix-related overrides, and also which Cabal it wants to see with GHC. `/nix/static-haskell-nix-newer-cabal-versions.patch` fixes Cabal errors that we didn't need because it couldn't apply a lack of patches that upstream already applied. `/nix/static.nix` is static build definitions. `/shell.nix` is for Nix users wanting a proper shell to work on Aura, with the right GHC & Stackage packages & all that.

fosskers:
I'm waiting to hear back from the guy behind the static nix stuff to see if Aura is already built by his Nix build servers. If so, we might be able to get these nice reduced binary releases "for free" without needing to commit anything to Aura directly.

bb010g:
It's not built right now. Aura is commented out due to build problems. https://github.com/nh2/static-haskell-nix/blob/dbce18f4808d27f6a51ce31585078b49c86bd2b5/survey/default.nix#L1319

Plus, we're on a LTS that upstream static-haskell-nix doesn't support due to Stack 2. This is why we're patching it here.

fosskers:
An update: https://github.com/NixOS/nixpkgs/issues/43795#issuecomment-652983539

fosskers:
Thank you for your work on this.

With some advice that I recently found [in the `stack` repo](https://github.com/commercialhaskell/stack/issues/3467), I was able just now to compile and release Aura with `-split-sections` turned on for all dependencies, reducing the final binary size down to ~9mb, just with `stack`. Keeping the project to just `stack` means much less maintenance for us, especially as the years go on. So I think it's best that we do releases this way from now on, even though it would mean foregoing your PR here.

Thoughts?

bb010g:
That sounds good. I'd still like the Nixpkgs support code for development, though, which is the first commit in here if you don't want me to make a new PR.

fosskers:
Yes that's reasonable :) Let me perform some git-foo.

fosskers:
I've cherry-picked the first commit from here. Thanks again for this.

PR #624, opened 2020-07-30
Attempt to parse config from XDG user config dir

samhh:
Hi! I'm the dude from Reddit. No Index yet, so here's a PR. :wink: 

The intention of this PR is to change the current Aura config behaviour from `try system | default` to `try user | try system | default`. The user config location is determined via [XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which is great for those of us who like to obsessively manage their [dotfiles](https://github.com/SamHH/dotfiles). A new dependency [xdg-basedir](https://github.com/sindresorhus/xdg-basedir) has been added for this purpose. A typical system's XDG config location would be `~/.config/aura/config`.

A behaviour I've observed, I think due to how the parsing works, is that an empty config file will be passed over, meaning that an empty user config will result in the `getAuraConf` function continuing onwards towards the system config. I'm unsure if this is desired behaviour.

I did consider creating a list of config locations and then trying them one by one, which'd be more extensible, but the last time I tried to utilise `Alternative` with `IO Maybe` it wasn't nearly as pleasant as just chaining a bunch of `<|>` together... don't know if it'd be feasible.

Not worked with Haskell professionally yet so let me know anything that needs changing or could be improved, however minor. Speaking of which, I'm unclear on best practices for dependency version ranges in Haskell, particularly for something that's 0.x. At the moment I've just pinned to `0.2.2` for stability.

Cheers!

### DIFF
--- a/aura/aura.cabal
+++ b/aura/aura.cabal
@@ -97,7 +97,7 @@ library
     , network-uri       ^>=2.6
     , stm               ^>=2.5
     , time              >=1.8 && <1.10
-    , xdg-basedir       ==0.2.2
+    , xdg-basedir       ^>=0.2
 
 executable aura
   import:         commons, libexec
--- a/aura/aura.cabal
+++ b/aura/aura.cabal
@@ -97,6 +97,7 @@ library
     , network-uri       ^>=2.6
     , stm               ^>=2.5
     , time              >=1.8 && <1.10
+    , xdg-basedir       ==0.2.2
 
 executable aura
   import:         commons, libexec
--- a/aura/lib/Aura/Settings/External.hs
+++ b/aura/lib/Aura/Settings/External.hs
@@ -14,7 +14,6 @@ module Aura.Settings.External
     AuraConfig(..)
   , getAuraConf
   , auraConfig
-  , defaultAuraConf
     -- * Parsing
   , Config(..)
   , config
@@ -23,11 +22,13 @@ module Aura.Settings.External
 import           Aura.Languages (langFromLocale)
 import           Aura.Settings
 import           Aura.Types
+import           Aura.Utils (hush)
 import           RIO hiding (some, try)
 import qualified RIO.ByteString as BS
 import           RIO.Directory
 import qualified RIO.Map as M
 import qualified RIO.Text as T
+import           System.Environment.XDG.BaseDir (getUserConfigFile)
 import           Text.Megaparsec hiding (single)
 import           Text.Megaparsec.Char
 import qualified Text.Megaparsec.Char.Lexer as L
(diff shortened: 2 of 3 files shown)


aura/lib/Aura/Settings/External.hs
```diff
@@ -45,17 +46,28 @@ data AuraConfig = AuraConfig
...
-      file <- decodeUtf8Lenient <$> BS.readFile fp
-      pure . either (const $ Config M.empty) id $ parse config "aura config" file
+      file <- decodeUtf8Lenient <$> BS.readFile path
+      pure . hush $ parse config "aura config" file
+
+getAuraConf :: IO Config
+getAuraConf = do
+  userCfg <- getAuraConfFrom =<< userAuraConfPath
+  case userCfg of
+    (Just x) -> pure x
```
fosskers, line 69: Parens are unnecessary here. Have you checked out [hlint](https://github.com/ndmitchell/hlint) yet?

aura/lib/Aura/Settings/External.hs
```diff
@@ -45,17 +46,28 @@ data AuraConfig = AuraConfig
...
     else do
-      file <- decodeUtf8Lenient <$> BS.readFile fp
-      pure . either (const $ Config M.empty) id $ parse config "aura config" file
+      file <- decodeUtf8Lenient <$> BS.readFile path
+      pure . hush $ parse config "aura config" file
+
+getAuraConf :: IO Config
+getAuraConf = do
+  userCfg <- getAuraConfFrom =<< userAuraConfPath
+  case userCfg of
```
fosskers, line 68: Since this `userCfg` symbol is only used in one place, do you want to try collapsing this with the previous line using [`LambdaCase`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=derivingvia#extension-LambdaCase)?

aura/aura.cabal
```diff
@@ -97,6 +97,7 @@ library
     , network-uri       ^>=2.6
     , stm               ^>=2.5
     , time              >=1.8 && <1.10
+    , xdg-basedir       ==0.2.2
```
fosskers, line 100: Despite not being updated since 2012, this is actually in Stackage. Neat. 

As for how to actually pin it, let's do `^>= 0.2`.

fosskers:
> A behaviour I've observed, I think due to how the parsing works, is that an empty config file will be passed over, meaning that an empty user config will result in the getAuraConf function continuing onwards towards the system config. I'm unsure if this is desired behaviour.

This is expected, yeah.

aura/lib/Aura/Settings/External.hs
```diff
@@ -45,17 +46,28 @@ data AuraConfig = AuraConfig
   , acAnalyse   :: !(Maybe BuildSwitch) }
   deriving stock (Show)
 
-defaultAuraConf :: FilePath
-defaultAuraConf = "/etc/aura.conf"
+userAuraConfPath :: IO FilePath
+userAuraConfPath = getUserConfigFile "aura" "config"
```
fosskers, line 50: Can we call the file itself `aura.conf`? This will mean fewer necessary changes for documentation elsewhere.

samhh: I've had a quick search and can't find anything on standardisation of filenames. I've noticed also that some other projects like mpd follow this naming convention (`app/app.conf`). :+1:

aura/lib/Aura/Settings/External.hs
samhh: I have. It works on my other Haskell code but everything in this repo gives me back "No hints" for some reason... something I need to investigate.

aura/lib/Aura/Settings/External.hs
samhh: Pushed a change, is that what you had in mind?

I've never thought of combining bind and lambdacase like this!

fosskers:
Exactly what I had in mind, looks great! Now I'll test.

aura/lib/Aura/Settings/External.hs
fosskers: Yes it's a nice trick that really cleans things up.

fosskers:
`sudo` throws a wrench into this, since the config path turns into `/root/.config/aura/aura.conf`. I'll fix it to specifically target `/home/<trueuser>`.

fosskers:
Fixed and merged. Thanks for this! I had to hard-code the path to be `/home/<theuser>/.config/aura/aura.conf`, which I realize might not be the true XDG dir, but the proper solution involves a greater issue that Aura has involving propagating the underlying user's Env to Aura when `sudo` is being used.

fosskers:
Do you use the default XGD dir? Will this work for you?

samhh:
Yep, I do, this serves my use case. Ideally to adhere to XDG it reads off of the environment variable so that it's configurable, but I think the vast majority of people use this default. :slightly_smiling_face:

fosskers:
Excellent, thanks. For now the hard-coding should be a safe default.

PR #653, opened 2020-10-22
Misleading 'check your input' Message

fosskers:
This PR (finally) gets rid of the infamous "Please check your input" message that would often appear in common cases.

### DIFF
--- a/aura/lib/Aura/Build.hs
+++ b/aura/lib/Aura/Build.hs
@@ -70,7 +70,7 @@ installPkgFiles files = do
 -- the user with flags.
 buildPackages :: NonEmpty Buildable -> RIO Env [PackagePath]
 buildPackages bs = mapMaybeA build (NEL.toList bs) >>= \case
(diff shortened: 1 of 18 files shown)


aura/lib/Aura/Build.hs
```diff
@@ -157,7 +157,7 @@ cloneRepo pkg usr = do
   logDebug $ "Currently in: " <> displayShow currDir
   scriptsDir <- liftIO $ chown usr currDir [] *> clone pkg
   case scriptsDir of
-    Nothing -> pure . Left . Failure . buildFail_7 $ bName pkg
+    Nothing -> pure . Left . Failure . FailMsg.  buildFail_7 $ bName pkg
```
ratijas, line 160: whitespace mis-click?

fosskers: Looks like it, although luckily that doesn't affect compilation.

PR #654, opened 2020-10-22
Wide Search

fosskers:
This PR implements a subflag for `-As` that allows it to yield results from the official repos as well.

### TODO

- [x] Confirm exit code details
- [x] Update manpage
- [x] Update CHANGELOG
- [x] Allow multiple search terms for `-As`

### DIFF
--- a/aur/CHANGELOG.md
+++ b/aur/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+#### Added
+
+- `Ord` and `Hashable` instances for `AurInfo`.
+
 ## 7.0.4 (2020-09-11)
 
 - Escape `+` that can occasionally appear in package names.
--- a/aur/aur.cabal
+++ b/aur/aur.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.2
 name:               aur
-version:            7.0.4
+version:            7.0.5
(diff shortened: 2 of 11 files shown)


ratijas:
So, ultimately you don't want to do that by default, instead requiring extra typing like subflag `-Ass` etc.?

fosskers:
That's right. In the end, `-A` is for AUR operations primarily. But this extra subflag can fulfill the needs of users like yourself who want both at once.

ratijas:
Idk. It's like both reasonable and not at the same time. What about separate Big Letter flag then?

These indented letters on the right are not used yet:
```
A
B
C
D
    E
    F
    G
h
    I
    J
    K
L
M
    N
O
P
Q
R
S
T
U
V
    W
    X
    Y
    Z
```

Let's say,
- it could be W (stands for Wide as you mentioned earlier) or X (for Cross-worlds), and these two have almost no intersection with lowercase options (only `-x, --unsuppress` for `-A`),
- I don't like Z because it is hard to reach on QWERTY keyboards, 
- Y — I can't think of any mnemonic, and it just doesn't sound right for me.
- N, K, J — same,
- l — don't ever use that one (oh no, it's ell, not big i, by mistake!)
- D, E, G — maybe, but why?

fosskers:
I've altered the long option to `--both`. Any thoughts?

PR #664, opened 2020-10-31
Update Fields.hs

oltulu:


fosskers:
Thanks for this :) I look forward to the final result.

fosskers:
Thank you!

PR #665, opened 2020-10-31
Update Languages.hs

oltulu:


fosskers:
Thanks, I can fuse the two PRs together manually.

fosskers:
Thank you!

PR #666, opened 2020-11-04
Bumps 'scheduler' upper bound

jkachmar:
Closes commercialhaskell/stackage#5720; I've verified that this compiles locally, but I haven't run the full `aura` test suite with this new version of `scheduler`.

### DIFF
--- a/aura/aura.cabal
+++ b/aura/aura.cabal
@@ -55,7 +55,7 @@ common libexec
     , http-client                  >=0.5 && <0.8
     , prettyprinter                >=1.2 && <1.8
     , prettyprinter-ansi-terminal  ^>=1.1
-    , scheduler                    >=1.1 && <1.5
+    , scheduler                    >=1.1 && <1.6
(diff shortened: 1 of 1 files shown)


jkachmar:
Question regarding the Nix integration: Is there an easy way to update the version of GHC that `nix-shell` pulls in?

When I updated `stack.yaml` to point to the latest Nightly snapshot (which uses GHC 8.10.2) and re-entered the `nix-shell`, it kept using GHC 8.8.4.

I ended up just downloading GHC 8.10.2 via `stack` to compile it, but I imagine there should be a way to do all this through Nix right?

jkachmar:
It also looks like the GitHub actions runner pulled this from cache (likely because it's using the Stackage LTS snapshot, so it didn't actually pick up the latest `scheduler` version).

fosskers:
Thanks Joe! I'll verify this quickly locally. The Nix config was added by someone else and is unmaintained by me (I have neither the means nor the knowledge).

PR #662, opened 2020-10-29
Rust Proof-of-Concept

fosskers:
This PR is not to be merged, it's meant strictly as a demonstration of techniques I plan to use in the real port.

See #657 .

Issue #657, opened 2020-10-23
Proposal: Port to Rust


fosskers:
With Aura nearing its 10th birthday, I'd like to open a discussion about the possibility of porting Aura to the [Rust] language. This has been on my mind for some time, but I don't believe that I'm being swept up in hype. Nor do I intend this discussion to be one of "Haskell vs Rust" in general. Both are fine languages. 

Such a change would represent Aura 4.0.0, and let me be clear that as far as users are concerned, they shouldn't notice any difference in Aura's commands / data formats / config formats. However when it comes to authorship and maintenance of Aura (which are primarily my personal burden), Rust as a systems language offers some distinct advantages.

### PROs

- **Access to `libalpm`:** This would improve performance in a number of places
  and unlock new features that were previously unfeasable. This is the form that
  I always wanted Aura to have, instead of relying on `pacman` so much.
- **More opportunity for outside contributions:** There are simply more Rust
  devs than Haskell devs.
- Better performance, smaller binaries, and ~static linking~ (we'll likely have to stay dynamic, due to `libalpm`).
- Official ecosystem support for release candidates. Overall easier to distribute.
- The ecosystem has much more momentum and will likely suffer less bitrot.
- Builds on more platforms/architectures (re: Macbooks will soon be using ARM
  chips. Related [stack issue]).
- Access to [clap] for CLI args:
  - **Can autogenerate completions.**
  - [Supports "fused short ops"](https://github.com/clap-rs/clap/blob/master/examples/23_flag_subcommands_pacman.rs) (that Pacman and Aura have) out of the box.
  - Offers suggestions if users make a typo.
  - Can define argument groups for the subflags of the big commands (`-S`, `-Q`,
    etc.). **This should fix the currently garbled `--help` output for free.**
- Error handling would be more straight-forward (re: IO exceptions in Haskell).
- Saner `String` handling.
- #572 #625 #640 would be fixed for free. 
- A rewrite acts as a "spring cleaning" of sorts, and would clear away historical cruft.
- Can get rid of the excess config.
- Aura would be versioned via SemVer instead of PVP.

### CONs

- I risk upsetting long-time Haskell users.
- Potential conflict with the [rua] project, although Aura is a drop-in
  replacement for Pacman, where `rua` is just for building AUR packages.
- Effort of porting the [language strings](https://github.com/fosskers/aura/blob/073740f60bfc1b3911c5312c2d96d624b8de84f5/aura/lib/Aura/Languages.hs).
[...]

PR #671, opened 2020-11-13
Rust Port: Part 1

fosskers:
This mega-PR ports Aura to Rust as per the discussion in #657 and the proof-of-concept in PR #662.

### Growth of Stripped Release Binary

Notes:
- Current Haskell-Aura is 8979kb.
- A number of unrelated commits usually exist between each measured growth point, so the increase isn't always due solely to the new dependency.

| Stage | Binary Size | Increase | Dependencies | 
|-|-|-|-|
| `clap` + `alpm` | 886kb | N/A | |
| `ubyte` | 911kb | +3% | |
| `rustyline` | 1114kb |  +22% | |
| `termcolor` | 1118kb | +0.4% | |
| `i18n-embed` | 1366kb | +22% | 98 |
| `-L` complete | 1394kb | +2% | 99 |
| `simplelog` | 1534kb | +10% | 99 |
| `rayon` | 1650kb | +8% | 109 | 
| `pbr` | 1678kb | +2% | 113 |
| `panic = "abort"` | 1538kb | -8% | 113? |
| `open` complete | 1558kb | +1% | 123 | 
| `pacmanconf` | 1590kb | +2% | 125 | 
| Error instances + no exec-lib | 1586kb | | 125 |
| `deps` complete | 1598kb | 0.7% | 127 |
| `itertools` | 1622kb | +2% | 128 |
| `-Cc`, `-Cy` and `curl` | 1726kb | | 137 |
| `-C` complete | 1774kb | | 140 | 
| Rust 2021 | 2229kb | |

Other notes:

- `versions` (pulling `nom`) only adds ~30kb.
- Dec 10: `opt-level = "s"` reduced final stripped size from `1710 -> 1410`

### Tracked Issues
- [x] https://github.com/kellpossible/cargo-i18n/issues/44
- [x] https://github.com/kellpossible/cargo-i18n/issues/45
- [ ] https://github.com/clap-rs/clap/discussions/2223
- [x] https://github.com/Morganamilo/pacmanconf.rs/issues/2
- [x] https://github.com/Morganamilo/alpm.rs/issues/12
- [ ] https://github.com/kellpossible/cargo-i18n/issues/52

### Tracked PRs

- [x] https://github.com/kellpossible/cargo-i18n/pull/46
- [x] https://github.com/kellpossible/cargo-i18n/pull/47
- [x] https://github.com/kellpossible/cargo-i18n/pull/53
- [x] https://github.com/Morganamilo/pacmanconf.rs/pull/4

### Before Release

- Reconsider `panic = abort`


Morganamilo:
Getting the command line working right is gonna be fun ;)

image

Morganamilo:
By the way, why is this split up into so many different creates?

`aura   aura-arch   aura-common   aura-core`

I do feel this is a bit unnecessary. You could easily have it all as one lib with different modules. Any reason you don't do this?

fosskers:
> Getting the command line working right is gonna be fun ;)

Heh, yes. Well even in current Haskell-based Aura, those flags must occur *after* the capital letter command. Nobody has complained about it up to this point, which either means it hasn't been an issue, or nobody has told me ;)

I think for the flag compat, as with the current implementation, I'm going with a best-effort approach. Will there likely remain some funky edge case that `pacman` would accept that Aura barfs on? Probably. Would any human actually try to use `pacman` that way? *Probably not*. And if they do, and have a serious issue with Aura's inability to handle it, then we can revisit then.

At the moment, this Rust port can already display usage info (i.e. `-Sh`) for each Pacman command identical to how Pacman itself displays it (modulo the options/flags sections that clap splits them into). This is already much better than what Haskell Aura does, so I'm thrilled.

> By the way, why is this split up into so many different creates?

This is by design. Haskell Aura is one-lib-one-exec, and many things were added to it organically over the years. There are often cross-cutting concerns and organizational problems that make the library hard to use. The Rust crate layout is still preliminary, but this would allow us to streamline:

1. Errors
2. Log messages
3. Localization
4. Dependencies

Then, if anyone ever needs a particular piece of Aura, they aren't bogged down by the concerns of the other pieces. Critically, `aura-arch` will be the only piece with a direct dependency on `alpm`. The others could act independently if they need to.

Morganamilo:
> Would any human actually try to use pacman that way? Probably not

I do -_-

While I get the idea of splitting up into so many crates. It seems a bit over board don't you think? Too many crates also becomes a bit of a pain to develop and maintain.

How about aura as the binary? This would contain everything aur specific. aur-lib as a general purpose aur lib?

There's also alpm-utils which I maintain. And is basically what you expect. Helper functions to make doing alpm stuff easier.

the general purpose aur-lib is also something I've wanted to do. The goal would be to split it out of paru and make depends on it.

So if you wanna work together to make it a common dependency I'd be all for that.

fosskers:
> I do -_-

:sweat_smile: By "edge case" I was imagining very strange arg orders + mixtures of long and short options + positional args. I'm confident we can arrive at something workable with `clap`. 

> Too many crates also becomes a bit of a pain to develop and maintain.

Yes this is definitely a concern. Of course this port is still in early days, so let me share with you my thought process. By the way, it's quite nice to be discussing these things with someone else who understands the whole scope of the problem :smile: 

My first step was to take [Aura's current layout](https://hackage.haskell.org/package/aura) and ask how this could be better organized, so that logic could be better shared with other tools/frontends/libraries/etc. Here's the first pass:

modules

I elaborate below. A number of these points are obvious when read, but they're that way because at some point they *weren't* (or still aren't!) obvious to Haskell Aura. So we have:

1. The CLI-based `aura` executable itself. 
  - All user interaction occurs here.
  - Providing a logging backend occurs here (whereas the libs use `log`)
  - All `fluent`-based localisation occurs here (but only English is baked into the executable).
  - The interpretation/handling of all errors from lower libs occurs here.
2. Some "core" library that handles common package manager tasks.
  - No user-visible IO occurs here. All results given as `Result<Foo, Error>`.
  - This library could be used by some other UI (ncurses? gtk?) that wants everything that Aura "does" without any of the specific CLI concerns or localization.
  - **It's true that some aspects of this would be Aura specific, while some would be applicable to any such AUR-compatible package manager.** Perhaps there's a better way to split them up without just shoving everything into the executable?
3. A layer over `alpm` concerned with core concerns of an Arch Linux system. 
  - Two simple examples already present in this PR: help with creating `Alpm` connections, and detecting orphan packages.
  - Could have any name, and doesn't need to be branded with "aura".
  - There is almost certainly some overlap here with `alpm-utils`.
4. Other modules which are pure Rust and need no `libalpm` binding.
5. Core types in Pure Rust common to multiple components.

I'm sure this will evolve as the port continues. Either way, I would be thrilled to collaborate with you on some common components. The more eyes we have on these problems the better - which is one of the reasons I celebrate that there are so many AUR tools, even if they're all quite similar and written in the same languages.

Morganamilo:
As far as a generic aur-lib goes. I can see maybe moving the [split_ functions](https://github.com/Morganamilo/paru/blob/e860dad5bbe1921aba478a7ca5c0d680278baa9f/src/util.rs#L44) and the comments/news grabbing,

Morganamilo:
Also worth noting the split functions need better names.

Morganamilo:
So I made a repo libaur and added the mentioned functions to it. If you want work on it together and add stuff please do.

https://github.com/Morganamilo/libaur.rs

fosskers:
Amazing, yes I will. So that repo is the place where we intend AUR-specific functionality (say, pulling a PKGBUILD) to live? It would contain things that any AUR tool would need?

Morganamilo:
> say, pulling a PKGBUILD

In theory, if I didn't already have https://github.com/Morganamilo/aur-fetch.rs :P

> what shouldn't it contain?

Anything to do with formatting, or forces formatting one way. If you look at the news code, I made a special effort to sill allow you to format it however you want.

fosskers:
> In theory, if I didn't already have https://github.com/Morganamilo/aur-fetch.rs :P

Heh, thank you, I may very well borrow that too. I wrote [the Haskell variants](https://hackage.haskell.org/package/aur) long ago, but can't exactly use them here.

Morganamilo:
Hmm, that lib seems to be for the aur-rpc. The rust lib for that is `raur`, `aur-fetch` is for cloning and diffing pkgbuilds
 .

fosskers:
Ah, gotcha. In that case the PKGBUILD fetching/diffing has always lived directly in Aura.

Morganamilo:
The library handles downloading concurrently, /diffing/marking as seen. There's quite a bit to it.

fosskers:
2020-11-18-170120_956x511_scrot

Morganamilo:
I'm still a fan of just using pacman. Alpm is usefully for dealing with packages, what is installed, what's available etc. But Using pacman for actual install/removes means you don't have to do the output formatting or worry about error handling and so on.

fosskers:
> I'm still a fan of just using pacman.

Me too, actually. This latest code is mostly practice, and there's a good chance I'll rip out the manual forming of a transaction and replace it with a call down to `pacman -Rsu`. So far the writing of this has informed me of:

- further `alpm` usage
- some aspects of module organization
- text colouring
- user input

I also plan to rip out `rustyline` today, I don't think it's useful here.

fosskers:
In general I think we should be careful about "stealing `pacman`'s thunder". If something is easier / safer for `pacman` to do, then `pacman` should do it. If some Aura-based process can be enhanced by `alpm`, then full-steam-ahead. If a traditionally pacman-based process can be significantly improved in some way via Rust, then we might consider it.

fosskers:
Actually I'll keep `rustyline` - it prevents issues with the arrow keys.

Morganamilo:
> Actually I'll keep `rustyline` - it prevents issues with the arrow keys and Ctrl+C, etc.

Honestly I prefer to keep those issues just to be consistent with Pacman.

fosskers:
Oh hah, I just noticed for the first time that Pacman freaks out when arrow keys are pressed too. I'm going to call that a bug on their end.

fosskers:
I've achieved the combination of localization + colouring, so we're pretty much full steam ahead. I also got good news today regarding the double flags: https://github.com/clap-rs/clap/discussions/2192#discussioncomment-133916

fosskers:
@Morganamilo So `clap` has been great for generating structs for all our flags, and the help output is wonderful, but do you have a suggestion for how to cleanly convert the parsed struct back into string args, for when `pacman` is called on the command line? You can see in `flags.rs` the "dumb" solution, namely the `ToArgs` trait and writing out a huge string of `if` statements. Haskell Aura has an easier time with this, since the Pacman flags are parsed a different way (but, help output is heavily sacrificied).

Is there some Rusty solution that comes to mind? Otherwise, iterating through a raw `ArgMatches` would be nice, but how exactly to do that is an [ongoing discussion](https://github.com/clap-rs/clap/issues/1206) among the clap folks.

Morganamilo:
What I do is define an array of pacman args. With clap you could then just iterate over them and call values_of on each flag.

Also I've neverr actually used clap v3 my attempts were with v2 .

Morganamilo:
Not sure if you have any use for this: https://github.com/Morganamilo/paru/blob/master/src/args.rs

fosskers:
A thought just occurred to me. If the `clap` parse succeeded, then by definition whatever is in the [`std::env::Args`](https://doc.rust-lang.org/std/env/struct.Args.html) iterator _is_ exactly what could be passed to `pacman`, as is. All I have to do is `collect()` on that Iterator!

fosskers:
At the moment I don't plan to release Aura v4 until `clap` gets out of beta. Either way, I do plan a beta period for Aura itself where people can try out the Rust version and report any issues before I officially scrap the old version.

Morganamilo:
> A thought just occurred to me. If the `clap` parse succeeded, then by definition whatever is in the [`std::env::Args`](https://doc.rust-lang.org/std/env/struct.Args.html) iterator _is_ exactly what could be passed to `pacman`, as is. All I have to do is `collect()` on that Iterator!

In paru at least there's a ton of extra flags. forr example `paru -Syu --devel` Does aura not do that?

fosskers:
All AUR-based functionality is put under an `-A` command. The pacman commands are left totally untouched. Some people don't like that they're separated, but it's one of the things that makes Aura what it is.

fosskers:
^ Works like a charm.

Morganamilo:
Ah,  I see, you're talking about standard pacman operations. So this wouldn't come in handy for things like `aura -Ayu --devel --debug`?

fosskers:
Unfortunately not, for those I have to form a unified `Env` type (yet unwritten) that gets passed around.

Morganamilo:
Is the Env type at all similar to https://github.com/Morganamilo/paru/blob/master/src/args.rs ?

fosskers:
Not quite, although similar to how the [Pacman flags all have inner structs](https://github.com/fosskers/aura/pull/671/files#diff-d5e75ed62c0d5702acbd05b3d6c818793cd275cf1d76297f48cb95c96b1cea37R71), the Aura ones will too. Instead of a unified `Env`, it might also be possible to just pass around that inner flag struct. I haven't yet decided.

Whatever type does eventually get passed around, it needs to know things like the build user, the build locations, everything about `pacman.conf`, etc etc. But these may only be `-A` concerns. The others, like `-B` and `-C` probably don't need to care.

Morganamilo:
Ah, well I have a huge config struct that just holds the state of everything. The above link is specifically about building a pacman command. And you don't really need to worry about the config there, because pacman will read the config itself.

fosskers:
> I have a huge config struct that just holds the state of everything.

Yeah the old `Env` type in Haskell Aura is exactly that.

fosskers:
I dropped `termcolor` for the much simpler `colored`.

Morganamilo:
Makepkg actually has 3 possible file locations:  `/etc/makepkg.conf`, `$XDG_CONFIG_HOME/pacman/makepkg.conf`, `~/.makepkg.conf`

fosskers:
Ah, thank you, I'll add checks for those.

fosskers:
`indicatif` is quite a heavy dep because it pulls `regex` among others, so I went with `pbr` instead.

Morganamilo:
I notice you're making a lot of vars like this? 

/// The default filepath of the Pacman/libalpm database.
pub const DEFAULT_DB: &str = "/var/lib/pacman/";

/// The default filepath of the package cache.
pub const DEFAULT_CACHE: &str = "/var/cache/pacman/pkg/";

/// The default filepath of the Pacman configuration.
pub const DEFAULT_PAC_CONF: &str = "/etc/pacman.conf";


You should probably just dep https://docs.rs/pacmanconf/0.1.1/pacmanconf/ and let that figure it out for ya.

fosskers:
Yes I totally plan to :)

fosskers:
The `aura_arch` subcrate is something of an experiment. While I'm developing I put things in there that seem appropriate, but the ultimate goal would be to phase it out completely, moving handy functions I discover to `alpm-utils`, etc.

Morganamilo:
Just saying, I will probably be pretty strict on what goes into alpm-utils. That's why i make libaur for more general stuff.

Morganamilo:
> Some(s) => s.ends_with(".pkg.tar.zst") || s.ends_with(".pkg.tar.xz"),

There's way more package extensions than that. I say you have two options. Filter on `.contains(.tar.)` which could have false positives. OR do `alpm.pkg_load(pkg).is_ok()` which is full proof but probably slower. No idea if it's noticeably slower though.

Morganamilo:
By the way I have multiple cache dirs, do you plan to handle that?

fosskers:
Thanks. As a first pass, I went with what I found in my own cache. As I suspected there are more patterns after all, eh. 

Checking how I detect built package tarballs after a `makepkg` call in Haskell Aura, I'm doing:
```haskell
  (_, out, _) <- readProcess $ proc "sudo" ["-u", T.unpack usr, makepkgCmd, "--packagelist"]
```
which frees me from needing to know the patterns ahead of time. Dealing with the cache is a different story though. For that, I'm doing:
```haskell
-- SILENT DROPS PATHS THAT DON'T PARSE
-- Maybe that's okay, since we don't know what non-package garbage files
-- could be sitting in the cache.
-- | Given every filepath contained in the package cache, form
-- the `Cache` type.
cache :: [PackagePath] -> Cache
cache = Cache . M.fromList . mapMaybe (\p -> (,p) <$> simplepkg p)

-- | Given a path to the package cache, yields its contents in a usable form.
cacheContents :: FilePath -> IO Cache
cacheContents pth = cache . mapMaybe (packagePath . (pth </>)) <$> listDirectory pth

-- | All packages from a given `Set` who have a copy in the cache.
pkgsInCache :: Settings -> Set PkgName -> IO (Set PkgName)
pkgsInCache ss ps = do
  c <- cacheContents . either id id . cachePathOf $ commonConfigOf ss
  pure . S.filter (`S.member` ps) . S.map spName . M.keysSet $ _cache c
```
So it grabs everything and tries to parse each filename via the `simplepkg` function. Looking there, it seems to be looking for an outdated extension:
```haskell
    post = char '-' *> (string "x86_64" <|> string "any") *> string ".pkg.tar.xz"
```
So good, we're dodging a bug here in the Rust.

Morganamilo:
Also are you going to merge this any time soon? To a separate branch of course. That would mean I can actually hack on it, instead of just being annoying and pointing out mistakes.

fosskers:
> By the way I have multiple cache dirs, do you plan to handle that?

Yes I do.

> That would mean I can actually hack on it

How about doing a PR that targets this branch? Mind you things are still rather fluid and liable to change. I'm circling around, completing the simpler commands first (`-C` now, next is `-B`) before tackling AUR packages. That way, the overall design should be informed before I even get there.

Morganamilo:
> By the way I have multiple cache dirs, do you plan to handle that?

Also to add to this. I'm quite the power user when it comes to pacman and friends. Everything on my system is weird and unexpected. Helps to break aur helpers that make assumptions on certain things :P

Morganamilo:
> How about doing a PR that targets this branch?

Hmm that would be a pr targeting a pr lol. Idk if I like that.

fosskers:
> Helps to break aur helpers

This is a good thing :smile: Another reason for doing this Rust port was to go over everything again and clear out all these old assumptions.

fosskers:
> Hmm that would be a pr targeting a pr lol. Idk if I like that.

I've done it before in industry, it's too bad to manage.

Morganamilo:
So I'm looking at improving the error handling. The thing is, the aura lib kind of confuses me. In my eyes, a library generally should not print output or take input from the user. This makes the aura lib feel like more of an application than a library.

Have you considered just deleting the lib.rs file and then moving all the mods to the main.rs?

The reason I bring this up is the error handling library "anyhow". It's quite a nice tool for handling errors, It is however only really suitable for applications. However the aura lib kinds of feels like an application so I'm not really sure if it should be used.

fosskers:
Thanks for looking into this. Definitely the module layout is not finalized, and neither is the error handling. Here's how I'm approaching both:

### Modules / Project Components

1. `aura` executable
2. `aura` lib: Intended to be imported by no other project than the executable.
3. `aura-core` lib: Contains internal logic which could in theory be useful to other projects. Makes no assumptions about errors, logging, or what upstream UI (the Aura exec, etc.) is even consuming it. Minimal dependencies.
4. `aura-arch` lib: A sort of wrapper/extension over `alpm`. As time goes on this seems less and less necessary and can hopefully be phased out, with its remaining logic moved to the `aura` exec or `alpm-utils`.

In general I'm not against moving all the current `mod` lines into `main` as you've suggested. `aura-core` should still remain its own separate crate though, for the dependency reasons I mentioned. Thoughts?

### Errors

Currently `main.rs` does nothing special with the errors it receives from function calls, it just dumbly prints them out. I think this is fine while we're still in a "churn" phase. It's useful for debugging. Haskell Aura obviously does the right thing, so sooner or later I will modify `main.rs` to emulate that. 

You're right to bring up `anyhow` - I've used in a number of projects. For the moment, I'm using and expanding a hand-made error type until it seems like I shouldn't. So far the amount of manual `.map_err(Error::Whatever)` calls I've had to do aren't many, and aren't yet an eyesore. We'll see how I feel in a few weeks.

Morganamilo:
Yeah it's fine for aura-core to be it's own lib (although sounds like it has the same goal as libaur), but I do see no reason for a lib/main split in aura itself.

The thing with `.map_err(Error::Whatever)` is you get no context. The errors you'll get are `File Exists` or `failed to initialise`. I generally just go with anyhow from the start because it's easier.

Seems as you're onboard with both i'll make it another PR :P

fosskers:
Let's hold back on `anyhow` for now, but I'm :+1: for combining the mods.

Morganamilo:
Any reason for that? The error handling is kind of awful right now and I'm happy to improve it. Either using anyhow or manually. Imo, it's good to get a good base going so when you finally decide to fix the errors you're not going through thousands of lines of code.

fosskers:
What would be the manual approach? I personally haven't found the current approach that painful to use.

Morganamilo:
Implement Error and display for the error stuct.

Implement `From<Io::Error> for Error` and other error types, this makes `?` just work and no need to map.

Add context to the error variants. So maybe instead of `Io(Io::Error)`  do `Io(String, Io::Error)`. then in the display implementation do `Io(file, err) => write!(fmt, "{}: {}", file, err`)`.

fosskers:
> Implement Error and display for the error stuct.

This I had planned to do, yes.

> Implement From<Io::Error> for Error and other error types, this makes ? just work and no need to map. 

Ah okay, that would be handy. Basically replacing what `anyhow` offers.

> Add context to the error variants. 

Yes this is possible. Might not be necessary for each variant, but for some it might be. Let's leave this out for now. I promise that the errors will be caught and handled properly in `main.rs` soon :smile:

Morganamilo:
> I have nothing against anyhow per se, but if we can get away with not adding a dependency, I'd prefer that. Another experiment I'm doing is seeing how small I can keep the final stripped binary.

Yeah Don't get your hopes up. Once you start doing AUR stuff you're going to need hyper, and thus tokio, Pulling in all the web and async stuff. They can be a bit beefy.

Paru is 6.1MB

fosskers:
`isahc` or raw `curl` are also possibilities.

Morganamilo:
If you use curl then you're losing out on async. Which is a really nice thing. Also curl's api is annoying. I would much prefer to use hyper and eat the size increase. What is 5MB vs 6MB anyway?

fosskers:
It's something of a challenge to me. I also enjoy minimalism in general. Here's an article I wrote about dependency reduction (although for Haskell): https://www.fosskers.ca/en/blog/wide-haskell

fosskers:
Heh, looks like dropping the exec lib saved 4kb :wink:

Morganamilo:
By the way, I must really recommend against panic=abort.  Even though a program shouldn't panic, it happens. And you want the use to be able to understand and report it.

fosskers:
Thanks, I've made a note for myself at the top of this thread.

Morganamilo:
Btw looking at the last commit. Do you have any intentions to use alpm-depends?

fosskers:
I do! The above commit is unrelated to dependency solving.

fosskers:
Thanks for the `Dep` lifetime fix:

graph

fosskers:
"Why the heck do I still have `python2` installed?"

```
cargo run -- deps python2 --reverse | dot -Tpng > graph.png
```
graph

Morganamilo:
I just use 'pacman -Qii python2' or 'pactree'

fosskers:
Yup, totally possible to use `pactree` for similar workflows. DOT is a pretty powerful format, so we'll be able to extend this more in the future for even prettier visualizations.

Morganamilo:
Pactree has --graph for dot output.

fosskers:
I'm quite pleased with this.
graph

Morganamilo:
7f50b9d could be written as:

```rust
use alpm_utils::DbExt;

fn foreigns(alpm: &Alpm) -> impl Iterator<Item=Package> {
    let dbs = alpm.syncdbs();
    alpm.localdb().pkgs().iter().filter(|p| dbs.pkg(p).is_err())
}
```

fosskers:
Oh nice usage of `find_map` in the internals! This is one of the functions I keep forgetting exists.

fosskers:
2020-12-04-180632_956x511_scrot

fosskers:
Pinging @atugard, re: a flag for specifying the privilege escalation command. That's actually something that could be added right away (in `flags.rs` in the `Args` struct) although we're not quite at the point where we could use it yet. Soon, though. I'm working as fast as I can.

fosskers:
Question for @Morganamilo regarding CI-based testing: do you have any recommendation for being able to compile this in a CI environment? The CI runners are Ubuntu by default, so installing some `pacman` package is probably not an option. Should I load up a docker container?

Morganamilo:
I'm not really sure. Gitlab's CI stuff lets you just choose the OS to be archlinux so I really want to do that. I even have a mirror at https://gitlab.archlinux.org/morganamilo/alpm.rs. But right now the public can't sign up to arch's gitlab instance so then no one would be able to make issues. I'll probably make the switch when they add support.

On the githib side, I've never used actions but I know yay has it working so maybe just copy that.

fosskers:
I'm a huge fan of GH Actions, so I'll take a look at what Yay does, thanks.

fosskers:
Interesting, `alpm` actually compiled, but `aura_arch` did not.

Morganamilo:
Yep alpm.rs will build fine without libalpm installed. It's when you try to do the linking in the executable you get the error.

fosskers:
Yup he's using an image: https://github.com/Jguer/go-alpm/blob/master/.github/workflows/test.yml#L18

fosskers:
Hot damn it works.

.github/workflows/rust.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Install Rust
+        run: pacman -Sy rust gcc --noconfirm
+
+      - name: Clone Project
+        uses: actions/checkout@v2
+
+      - name: Cache Dependencies
+        uses: Swatinem/rust-cache@v1
+        with:
+          working-directory: rust/
```
fosskers, line 27: TODO This will need to be changed once I make the Rust the main code base!

Morganamilo:
> [ci] Rust needs `openssl`

I don't get this, Pacman does the dep solving for you right?

fosskers:
Not for dep solving - it's for the `curl` lib, which needs to find to the system's `openssl`.

Morganamilo:
Right. I'll say the commit is a bit misleading "rust needs openssl" on the same line where you're `pacman -S rust`.

Morganamilo:
I will also say it's a bit weird to use curl instead of hyper. I assume you'll eventually dep raur right? That uses hyper, although if you want to add it a curl backend I wouldn't be against it.

fosskers:
I plan to not include anything from Async Rust, in order to keep complexity and binary size down.

Morganamilo:
Async really is great though. Personally I love it.

fosskers:
Oh I have nothing against `async` in general, I've used it with success elsewhere. In Aura's case, I think I can get away with just `rayon`.

leshow:
Did you evaluate any other cli arg parsers other than clap? It can really stretch compile times and binary sizes.

I've used [argh](https://github.com/google/argh) to good effect before, and it has lightning fast compile times. There's also [pico-args](https://github.com/RazrFalcon/pico-args). Not sure if either of these covers all the cases aura needs however

fosskers:
Usually I use `gumdrop`, but `clap` in this case offer a number of special powers that I have found elsewhere. See [the original port proposal](https://github.com/fosskers/aura/discussions/657) for more info.

fosskers:
Consider https://lib.rs/crates/solvent .

fosskers:
@Morganamilo How much effort would it be (for me) to deasyncifize `raur`? Say by providing a `curl` backend via feature macros? Or does the existing API thoroughly assume `async`?

Morganamilo:
There's a sync and an async API. What needs to be done is separate the traits from the implementation (I have a branch that does that already)  Then you can use the trait and you can just impl it using curl.

Though the aur-depends crate does requite it to be async  is you were wanting to use that.

fosskers:
> Then you can use the trait and you can just impl it using curl.

Nice! I'd be happy to upstream that.

> Though the aur-depends crate does requite it to be async is you were wanting to use that.

I'll have to roll my own there, then. Won't be rocket surgery, done it once in Haskell already.

Morganamilo:
See https://gitlab.com/davidbittner/raur/-/merge_requests/22

Once merged you can use `--no-default-features.blocking-trait` then `impl raur::bloicking::Raur<Error> for MyHandle>`. You'll have to source raur from the git repo for now. I'll let you test it first before pushing a new version.

Morganamilo:
> > Though the aur-depends crate does requite it to be async is you were wanting to use that.

> I'll have to roll my own there, then. Won't be rocket surgery, done it once in Haskell already.

Funny thing is it doesn't require async. It just uses the async raur so it too has to be async. Would be nice if you could abstract over being async but you can't.

Morganamilo:
I managed to get aur-depends non async and generic over the error type so you should be able to use it. The new code to do so is in the `blocking` branch of my repo.

fosskers:
`-Ai` done! Feels good.

Morganamilo:
stdout is a mutex internally. So when printing many times it's more performant to lock that mutex and then call write manually. However I don't really care. The real world performance is not really impacted.

Morganamilo:
> [aura] Configurable clone path

Consider also supporting the env var `AURDEST` for this. Supported by, pacaur, aurutils, yay and paru.

fosskers:
Great idea. Are there expectations there w.r.t. the worktree of each clone being clean? Do tools expect to build in there too?

Morganamilo:
Just clone to there. Whatever else you want to do is then up to you. I personally use -Cc to clean before/after.

fosskers:
The current clone dir is meant to stay clean, with builds occurring elsewhere, which sounds to be fine. Great then, I'll add support for `AURDEST` today.

Morganamilo:
Is this stalled? Just lack of time or are there any blockers?

fosskers:
It's still happening! I was out of the country recently and otherwise busy with work, but I'm returning to it very soon.

Morganamilo:
Good to hear, was just checking in.

fosskers:
And we're back!

fosskers:
@Morganamilo multiple caches supported now :)

Morganamilo:
So looking at the latest commits. Quick reminder I have libs for downloading aur packages and dep solving. They are async based but I did make versions that were optionally async.

It looks like II may have never pushed the branches.

Let me know if you're interested in the APIs.

https://github.com/Morganamilo/aur-fetch.rs
https://github.com/Morganamilo/aur-depends

fosskers:
Thanks I'll take a look! Yes keeping `async` out of things is important for me.

fosskers:
@Morganamilo when you have a chance, also please let me know about https://github.com/Morganamilo/pacmanconf.rs/pull/4. If that's merged, I can release `r2d2-alpm` generally, and then I plan to merge this branch here into `master` and continue the work in new, smaller branches.

Morganamilo:
By the way i don't get what r2d2-alpm is for. Alpm has parallel downloads itself.

fosskers:
It's not for package downloads, it's for when you want multiple `Alpm` handles within a multithreaded computation like a `map` (etc.) from a `ParallelIterator` from `rayon`.

fosskers:
https://crates.io/crates/r2d2-alpm

Morganamilo:
Hmm okay. Is that actually needed though?

Pacman tends to cache everything, so multiple handles means more cache misses and more time spent loading from disk. Plus the synchronisation overhead and all that.

Maybe you used to use this in aura which uses the pacman command? But when using alpm I don't see it making much sense.

fosskers:
Perhaps the amount of concurrency I'm doing won't pay off in the end. I'm measuring overall resolution times, and will write a strictly serial version afterwards to see how it compares. If it's just as good, then I can run off a single `Alpm` or look into `aur-depends` (which I haven't forgotten about).

PR #676, opened 2020-11-30
Add git feature

Morganamilo:
This allows building aura against pacman-git.

This feature can automatically be endabled by the pkgbuild when there is
one.

### DIFF
--- a/rust/aura/Cargo.toml
+++ b/rust/aura/Cargo.toml
@@ -23,3 +23,7 @@ simplelog = "0.8"
 ubyte = "0.10"
 unic-langid = { version = "0.9", features = ["macros"] }
 webbrowser = "0.5"
+
+[features]
+git = ["alpm/git"]
+


fosskers:
Looks good! Can this be merged as-is? For now I'm going to continue dev'ing on the Pacman/alpm I have natively installed, but I can see why you'd want the power-user setup :smile: Who knows, I may switch to this soon to test out the concurrent downloads.

Morganamilo:
Yes, it just works as it. There are some api changes for pacman-git but  aura doesn't use any of that so no other changes needed.

Also thought you would be using pacman-git seems as you sent a patch :P

fosskers:
> Also thought you would be using pacman-git seems as you sent a patch :P

I have a clone of it which I test with sometimes. I noticed the issue with `-Fh` when I was writing `flags.rs`. I'd use `pacman-git` more regulardly if I could figure out the concurrent downloading. Did the recent bugs get ironed out?

fosskers:
Thanks for your first patch to Aura :partying_face:

Morganamilo:
> I noticed the issue with -Fh when I was writing flags.rs

Funny thing is I may have caused that... I rewrote the -F UI a while ago.

> Did the recent bugs get ironed out?

Yeah, I patched 6 of them >.> looks good to me ATM.

> Thanks for your first patch to Aura 🥳

That's 9 aur helpers now :p

PR #677, opened 2020-11-30
Fixups and convert to anyhow

Morganamilo:
I went along with the anyhow thing to try and convince you :P Error messages are now much more readable.

If you don't like it I'll just pop the last commit.

before: `Error: Alpm(NotADir)`

after: `error: failed to initialize alpm: root='/' dbpath='fcedew': could not find or read directory`


rust/aura/src/main.rs
```diff
@@ -32,14 +32,12 @@ fn main() -> Result<(), Error> {
     // Establish common file paths.
     let logp: &Path = args
         .logfile
-        .as_ref()
-        .map(|p| p.as_path())
+        .as_deref()
```
fosskers, line 68: I will remember this.

fosskers:
I may need to forego the `anyhow` usage for now, considering that I will want all user-facing messages to be localized. Another reason for having the explicit variant wrappers around each error type. 

I can merge this and then hand-write the `Error` trait conversion code so that things are easier for you in the meantime.

fosskers:
And double-checking the way docs render with the lib now gone, everything looks as expected :+1: I frequently utilize my own projects' docs while developing so that was critical to me.

fosskers:
The commits are now included back in the `colin/rust` branch, although Github seems not to have detected that here, so I'll close manually.

rust/aura/src/main.rs
ratijas: Oh, come on, just throw `&` and `*` at it — as many as needed, until it compiles.

PR #699, opened 2021-03-09
Escape ' in descriptions in zsh completion

amosonn:
Un-escaped `'` in some descriptions in zsh completion
were breaking it


fosskers:
Thanks, I was going to fix this today but you beat me to it :)

PR #705, opened 2021-05-08
Add raur-curl

Morganamilo:
raur-curl is an implementation of the raur API using curl as a back end.
It's pretty self contained so I've put it in its own crate.

This depends on the git version of raur. Once it's had some use here and
deemed suitable I'll push a new vesion of raur.


fosskers:
Hot damn! Thanks  a  lot for this, I'll get testing it tomorrow.

fosskers:
Thanks again, I'll be using this right away.

PR #710, opened 2021-05-19
Don't do multiple AUR requests for search

Morganamilo:
Instead of doing multiple requests and filtering the results into one
list, you can achieve the same by just searching using the first term
and filtering the results by the other terms.

This ends up being faster and puts less strain on the AUR.


fosskers:
Thanks, good eye. I'll incorporate in today.

fosskers:
Thanks!

PR #735, opened 2021-11-07
Add Romanian localisation

skrrt:
Here's a Romanian translation for Aura courtesy of myself and another contributor. Figured it might be worth it.


fosskers:
This is fantastic! Let me try it out.

fosskers:
Thanks again! It'll be included in the next release.

PR #746, opened 2022-01-17
Expand Norwegian translations

Areskiko:
Adds some new Norwegian translations. Some of these are the same as the English ones, though I think it's best to be explicit.


fosskers:
Thanks!

PR #747, opened 2022-02-05
Add Korean localization

ipegte93:


fosskers:
Finally someone has added Korean! Thank you very much for this! I will review it on Monday.

fosskers (changes requested):
Please also add translations to the `aura/lib/Aura/Languages.Fields.hs` file, and then I can merge this.

fosskers (approved):

fosskers:
Thank you very much!

PR #750, opened 2022-02-17
Rust Port: Part 2 - Dependencies

fosskers:
This PR expands on #671 by rounding off much of the internals of `-A`.

This is a vast improvement over the `Dependencies.hs` Haskell version in both performance and accuracy.


Morganamilo:
> say compared to what the Haskell does

What does haskell do?

fosskers:
I meant the Haskell version of dep resolution found here: https://github.com/fosskers/aura/blob/master/aura/lib/Aura/Dependencies.hs

fosskers:
And am I right that the AUR site itself support multiple search terms, but the RPC does not?

fosskers:
[This patch](https://lists.archlinux.org/pipermail/aur-dev/2020-July/004937.html) got my hopes up, but I don't see an actual v6 supported here https://gitlab.archlinux.org/archlinux/aurweb/-/blob/master/doc/rpc.txt

Morganamilo:
I've given up on searching for provides because I find my hack to be sufficient to be honest. Do note though that I only do the provider search for targets and missing packages. Doing it at every level is not ideal.

Morganamilo:
Also I pushed a threaded version of aur-fetch using crossbeam https://github.com/Morganamilo/aur-fetch.rs/blob/master/src/fetch.rs#L121

fosskers:
This is now comically fast and handles `provides` correctly.

Morganamilo:
Running your own aur mirror? Seems a bit fragile.

fosskers:
It's just a data mirror, luckily. Otherwise yes, that would be a nightmare.

Morganamilo:
How does it work as a speedup then? Can it handle the traffic of many users?

fosskers:
Because it doesn't use a database and holds all package data in memory, using Rust ref tricks to store `HashMap`s of `&Package` data for each kind of lookup you'd want to do (info, search, and provides).

Issue #758, opened 2022-04-17
"Failed to 'git pull' the latest updates" with VCS packages


mi544:
I have encountered an error when trying to install `protonup-git` from the AUR with aura  
```
[mi@unicorn ~]$ aura --version

 __ _ _  _ _ _ __ _    Pacman v6.0.1 - libalpm v13.0.1
/ _` | || | '_/ _` |   Copyright (C) 2006-2021 Pacman Development Team
\__,_|\_,_|_| \__,_|   Copyright (C) 2002-2006 Judd Vinet
AURA Version 3.2.7
 by Colin Woodbury     This program may be freely redistributed under
                       the terms of the GNU General Public License.
Aura Translators:
 (Japanese) Onoue Takuro / Colin Woodbury
 (Polish) Chris Warrick / Michał Kurek
 (Croatian) Denis Kasak / "stranac"
 (Swedish) Fredrik Haikarainen / Daniel Beecham
 (German) Lukas Niederbremer / Jonas Platte
 (Spanish) Alejandro Gómez / Sergio Conde / Max Ferrer
 (Portuguese) Henry Kupty / Thiago Perrotta / Wagner Amaral
 (French) Ma Jiehong / Fabien Dubosson
 (Russian) Kyrylo Silin / Alexey Kotlyarov
 (Italian) Bob Valantin / Cristian Tentella
 (Serbian) Filip Brcic
 (Norwegian) "chinatsun"
 (Indonesia) "pak tua Greg"
 (Chinese) Kai Zhang
 (Esperanto) Zachary Matthews
 (Dutch) Joris Blanken
 (Turkish) Cihan Alkan
 (Arabic) "Array in a Matrix"
 (Ukrainian) Andriy Cherniy
 (Romanian) 90 / benone
 (Vietnamese) "Kritiqual"
 (Czech) Daniel Rosel

[mi@unicorn ~]$ sudo aura -Ax protonup-git
aura >>= Determining dependencies...
aura >>= AUR Packages:
protonup-git
aura >>= Continue? [Y/n]
aura >>= Building protonup-git...
aura >>= Failed to 'git pull' the latest updates.
aura >>= Would you like to continue anyway? [Y/n]
aura >>= Every package failed to build.
```

I found a closed issue that referenced the same error (#615) so I tried the suggested command but the installation still failed  
```
[mi@unicorn ~]$ sudo aura -Ax --hotedit --force protonup-git
aura >>= Determining dependencies...
aura >>= AUR Packages:
[...]

PR #760, opened 2022-04-23
setuid and setgid with given user for `git pull`

ryandv:
Resolves https://github.com/fosskers/aura/issues/758

Confirming that aura built from this revision can successfully pull `mesa-git` (the one I was struggling with).

This is quick and dirty to get the thing working, so let me know if anything needs to be touched up.

### DIFF
--- a/aura/aura.cabal
+++ b/aura/aura.cabal
@@ -58,6 +58,7 @@ common libexec
     , scheduler                    >=1.1 && <2.1
     , transformers                 ^>=0.5
     , typed-process                ^>=0.2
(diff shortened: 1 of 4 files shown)


fosskers:
Thank you for this - it's likely cleaner that `sudo`ing down into the build user as I was going to do.

Note for anyone watching: this will not be an issue with Aura 4.

fosskers:
This doesn't work for me, so I'm continuing to investigate.

fosskers:
Thanks again!

PR #777, opened 2022-06-29
Auto-generated Man Pages

fosskers:
This PR leverages `clap` to auto-generate man pages for Aura. They will never need to be updated by hand again.


fosskers:
Salute to the man who put a lot of work into updating the manpage! @ratijas Your contributions shall live forever in the holy git log.

PR #787, opened 2022-10-17
Add missing Functor instance for These

Topsii:
Adresses haskell/core-libraries-committee#91


fosskers:
Thanks for going through the effort of the impact assessment!

PR #816, opened 2023-10-06
feat: adds hindi translation

yozachar:
:india:


yozachar:
Editing `aura/lib/Aura/Languages/Fields.hs` need to be mentioned in the [docs](https://fosskers.github.io/aura/localisation.html).

fosskers:
Finally a Hindi translation! Thank you!

PR #818, opened 2023-10-22
Specify `--stack-yaml` in the installation instructions

sangaline:
Following the installation instructions from the README did not work for me because stack was not picking up the `stack.yaml` file in the `haskell` subdirectory. I needed to manually specify the location using the `--stack-yaml` argument.


fosskers:
Thanks! I added an extra commit to mention the option of installing the Rust-based Alpha as well.

PR #843, opened 2024-06-11
Support for `pkgctl build`

### LINKED ISSUE #838: Add option to use `pkgctl build` instead of `makepkg` or similar
Bit of an enhancement thing, more than a few times you'll run into some AUR package that gets _really_ finicky with regards to what's on your system. For those cases, or just in general for better sandboxing, it's really convenient to be able to use the relatively new system of `pkgctl build` to automatically create a proper sandbox within which solely the dependencies for the build get installed, and build the package there.

Simply having an option to use `pkgctl` would be a nice improvement for those edgecases, even better if it's able to store a configuration option saying "hey package `foo` needs pkgctl" to use during the next `aura -Akuax` operation.

If `aura` could go further and also handle more complicated cases like chained AUR dependencies, that could be really cool too. I.e. if AUR package `foo` depends upon package `bar` and it's available both in the main repos and the AUR (or only in the AUR), as far as I know you need to explicitly include the already built package `bar` to `pkgctl build` with the `-I...` option. So if this hypothetical option were able to make `aura` build `bar` first with `pkgctl` and then `foo` with `bar` included that would be really cool.

fosskers:
This feature is used to specify certain packages that the user wishes to be built via `pkgctl build`. By default this is not activated for all packages as setting up chroots can take some time, and local `makepkg`-based building is still easier to inspect.

Still, building via chroots is probably safer overall and should be the direction we move toward.

Closes #838


rust/aura-core/Cargo.toml
```diff
@@ -27,4 +27,5 @@ serde_json = "1.0"
 srcinfo = "1.1"
 time = { version = "0.3", features = ["serde", "local-offset"] }
 validated = { version = "0.4", features = ["rayon"] }
+versions = { git = "ssh://git@github.com/fosskers/rs-versions.git", rev = "cb2c74823434fe0dc3e7ae09fe1a33bb255c19d3" }
```
bb010g, line 30: If you use [url.\<base\>.pushInsteadOf](<https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtpushInsteadOf>), you can avoid publishing SSH remote URLs that require authentication.
```gitconfig
[url "ssh://git@github.com/"]
  pushInsteadOf = https://github.com/
```

fosskers: That's just temporary to test a minor change within `versions`.

PR #930, opened 2024-09-12
Add Bengali (বাংলা) 

saifshahriar:
I don't have any experience with rust. :(
Pleases verify my changes to the rust codes.
Also, I wanted to add two aliases for the language 'Bengali': `'bengali', 'বাংলা'`. Since I couldn't find a way to do so, I kept only the first one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
     :czech_republic:
     :kr:
     :india:
+    :bangladesh:
   </div>
   <br>
   
@@ -335,6 +336,7 @@ Aura has been translated by these generous people:
 | Language   | Translators                                     |
 |------------|-------------------------------------------------|
 | Arabic     | "Array in a Matrix"                             |
+| Bengali    | Saif Shahriar                                   |
 | Chinese    | Kai Zhang and Alex3236                          |
 | Croatian   | Denis Kasak and "stranac"                       |
 | Czech      | Daniel Rosel                                    |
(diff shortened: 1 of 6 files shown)


fosskers:
Fantastic, thanks a lot for this. I will install a Bengali font and then test this :laughing:

saifshahriar:
Thank you.
Kindly, try it inside **Konsole** terminal as it has the least worst support for Bengali fonts.
Please try them with these settings enabled.

Screenshot_20240913_110126



Here, how it might look with the font [Hind Siliguri](https://fonts.google.com/specimen/Hind+Siliguri) in Konsole:

Screenshot_20240913_105651

Here is how it might look with a Sans version of a Bengali Monospace font Mitra (not Mitra Mono. The monospace variant does not render spaces properly around Bengali fonts):

Screenshot_20240913_110206




Both has a good vowel support for Bengali with the Brahmic Script support turned on. But the spacing is butchered.




Here is how it might look in a minimal terminal like st (JetBrainsMono + Mitra Mono):

Screenshot_20240913_110459


It has good spacing, but no support for Brahmic Script, so the vowels are out of place.

Bengali script is kind of complex, even more than Devanagri/Hindi or Arabic. So, the devs are not interested. Maybe someday we might have an Aura GUI or properly supported terminals.


This is the best font that I have gotten to work with Konsole so far. I don't no how I have this variant in my system. You can find the font [here](https://www.omicronlab.com/bangla-fonts.html), not sure about the exact variant tough.

Screenshot_20240913_113620

fosskers:
Thanks for those screenshots. I'm going to try it on https://wezfurlong.org/wezterm/ , the terminal I've been using recently.

fosskers:
bengali

Looks good to me.

fosskers:
Thanks again!

PR #934, opened 2024-09-16
fix(-A): detect more compression formats

190n:
109ce6763e79e8a4123a6f9c5d7840ac2f536921 added support for uncompressed package tarballs, but the compression format was limited to xz or zstd. I added detection for all the compression formats mentioned in the default `makepkg.conf`:

```
COMPRESSGZ=(gzip -c -f -n)
COMPRESSBZ2=(bzip2 -c -f)
COMPRESSXZ=(xz -c -z -)
COMPRESSZST=(zstd -c -T0 -)
COMPRESSLRZ=(lrzip -q)
COMPRESSLZO=(lzop -q)
COMPRESSZ=(compress -c -f)
COMPRESSLZ4=(lz4 -q)
COMPRESSLZ=(lzip -c -f)
```

In particular this restores LZ4 support which broke for me when I recently upgraded to aura 4.0.

### DIFF
--- a/rust/aura-core/src/cache.rs
+++ b/rust/aura-core/src/cache.rs
@@ -274,6 +274,7 @@ pub fn is_package(path: &Path) -> bool {
                 ".pkg.tar.Z",
                 ".pkg.tar.lz4",
                 ".pkg.tar.lz",
+                ".pkg.tar",
             ]
             .iter()
             .any(|ext| p.ends_with(ext))
--- a/rust/aura-core/src/cache.rs
+++ b/rust/aura-core/src/cache.rs
@@ -264,7 +264,19 @@ pub fn missing_tarballs<'a>(
 pub fn is_package(path: &Path) -> bool {
     path.to_str()
         .map(|p| {
-            p.ends_with(".pkg.tar.zst") || p.ends_with(".pkg.tar.xz") || p.ends_with(".pkg.tar")
+            [
+                ".pkg.tar.gz",
+                ".pkg.tar.bz2",
+                ".pkg.tar.xz",
+                ".pkg.tar.zst",
+                ".pkg.tar.lrz",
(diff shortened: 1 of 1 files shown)


fosskers:
Hi, thanks for this. I will do a full review in 5 days.

rust/aura-core/src/cache.rs
```diff
@@ -264,7 +264,20 @@ pub fn missing_tarballs<'a>(
...
+                ".pkg.tar.zst",
+                ".pkg.tar.lrz",
+                ".pkg.tar.lzo",
+                ".pkg.tar.Z",
+                ".pkg.tar.lz4",
+                ".pkg.tar.lz",
+                ".pkg.tar",
+            ]
+            .iter()
+            .any(|ext| p.ends_with(ext))
```
fosskers, line 280: This function is used in a number of places and in some cases may be called thousands of times. I'd like to avoid the allocation of an `Iter` and the `O(n)` lookup if possible, but for that we'd need a `HashSet` or something, yet can't allocate one statically.

Another approach would be to just have a large chain of `||` calls, with the most likely candidates first. Any thoughts?

190n: I checked on Godbolt and all the iterator stuff seems to be inlined. [Here's the code generated with `any()`](https://godbolt.org/z/nE99Efjb6) and [here's the code with a `||` chain](https://godbolt.org/z/1zYxP7WT3). The order is a little different but the basic operations are the same.

I can move `.pkg.tar`, `.pkg.tar.zst`, and `.pkg.tar.xz` to the top, though. That does affect the order in which comparisons are made in the assembly so it'll be a bit faster to return early on those cases.

fosskers:
I took care of it. Thanks again for this.

PR #961, opened 2024-11-28
Russian translate and fix for Fluent plural handling

SuperSuslik312:
Fix Fluent plural handling for `$days` and retain red color formatting.

- Pass `$days` as a number to Fluent for correct pluralization.
- Replace the number in the localized string with a red-colored version
  using the `.red()` method.
- Ensures both proper plural handling and terminal color formatting.

These changes were implemented due to the specifics of plural forms in the Russian language.


fosskers:
Thanks!

PR #965, opened 2025-01-03
Add Latin

semilin:
Nobody asked for it, but here it is. I was highly amused when I went to the localisation page and saw that Latin was the example used, probably with the expectation that no one would actually implement it. Sorry! :smiley_cat:

This was a very enjoyable project. It's fun to translate things into Latin because the lack of modern vocabulary allows a lot of creative liberty. I tried to mostly avoid neologisms and semantic loaning cognate words, but there are a few words I conceded that preference on, e.g. I invented "cachea" for "cache" because I feel that's a somewhat important word to communicate clearly, and I used the existing "commissio" for git commits. I'm pretty happy with the other technical terms: "charta", literally "piece of paper" for "file"; "receptacula", literally "receptacle" or "container" for "directory"; most importantly "sarcina", which does literally mean "package"; and conversely, "requirenda", literally "that which is required", for "dependency". Even more than these my boldest translation might actually be "accessibilitas" for "permissions", which I honestly think captures the point of *nix file permissions better than the English, which misleadingly implies that the file itself has permissions to do something.

I think I've translated this in such a way that someone with intermediate Latin knowledge should have very little trouble understanding it, even with the necessarily unorthodox vocabulary. Not that this of any consequence, since it wouldn't surprise me if the number of decent Latin speakers who use Arch is exactly one, but it was fun to hold myself to the standard anyway!


fosskers:
What fortuitous timing - I'm starting my own Latin study tomorrow! 

Thank you for this.

fosskers:
Thanks again! Italia in Eurōpā est!

semilin:
Et Rōma in Italiā est! Hope you enjoy Familia Romana. Feel free to contact me if you want any help or just want to talk about Latin :smiley_cat:

Issue #996, opened 2025-08-23
The clean option doesn't clean the makepkg BUILDDIR


dtodorov512:
Steps to reproduce:

1. Set `clean = true` in `~/.config/aura/config.toml`
2. Set `BUILDDIR=/tmp/makepkg` in `/etc/makepkg.conf`
3. Build a package.
4. The directory at `/tmp/makepkg/<pkg_name>` will contain a file.

The fix is simple enough. Pass the `-c` option to makepkg when clean is set in the configuration file.

dtodorov512:
Done

PR #997, opened 2025-09-03
Running makepkg with -c when the clean configuration option is enabled.

dtodorov512:
Fix for issue #996 

Let me know if I have to add any tests or anything.


fosskers:
Thank you for this, I will get to it as soon as I'm able.

fosskers:
Thanks!

Issue #1008, opened 2025-12-13
AUR: `aura-bin` `libalpm` dependency mismatch


make-github-pseudonymous-again:
```sh
$ pacman -F libalpm.so.16
core/pacman 7.1.0.r7.gb9f7d4a-1 [installed]
    usr/lib/libalpm.so.16
```

```sh
$ pacman -Q aura
aura-bin 4.0.8-1
```

```sh
$ aura -Syu
aura: error while loading shared libraries: libalpm.so.15: cannot open shared object file: No such file or directory
```

Issue #1007, opened 2025-12-13
Aura does not build anymore due to `libalpm` dependency mismatch


osingaatje:
```
==> Making package: aura 4.0.8-1 (Sat 13 Dec 2025 14:25:17 CET)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Downloading aura-4.0.8.tar.gz...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0   0     0   0     0     0     0  --:--:-- --:--:-- --:--:--     0
100  4536k   0  4536k   0     0  4439k     0  --:--:--  0:00:01 --:--:-- 12046k
==> Validating source files with sha256sums...
    aura-4.0.8.tar.gz ... Passed
==> Extracting sources...
  -> Extracting aura-4.0.8.tar.gz with bsdtar
==> Starting prepare()...
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/douwe/aura/src/aura-4.0.8/rust/aura-pm/Cargo.toml
workspace: /home/douwe/aura/src/aura-4.0.8/rust/Cargo.toml
==> Starting build()...
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/douwe/aura/src/aura-4.0.8/rust/aura-pm/Cargo.toml
workspace: /home/douwe/aura/src/aura-4.0.8/rust/Cargo.toml
   Compiling proc-macro2 v1.0.86
   Compiling unicode-ident v1.0.13
   Compiling version_check v0.9.5
   Compiling libc v0.2.159
   Compiling serde v1.0.210
   Compiling typenum v1.17.0
   Compiling pkg-config v0.3.31
   Compiling autocfg v1.4.0
   Compiling thiserror v1.0.64
   Compiling parking_lot_core v0.9.10
   Compiling shlex v1.3.0
   Compiling crossbeam-utils v0.8.20
   Compiling proc-macro-hack v0.5.20+deprecated
   Compiling cfg-if v1.0.0
   Compiling vcpkg v0.2.15
   Compiling smallvec v1.13.2
   Compiling cpufeatures v0.2.14
   Compiling bitflags v2.6.0
   Compiling cc v1.1.22
   Compiling same-file v1.0.6
   Compiling scopeguard v1.2.0
   Compiling walkdir v2.5.0
[...]

PR #1009, opened 2025-12-14
Fixes libalpm mismatch (#1007 and #1008)

hbritto:
This PR fixes the libalpm.so.15 mismatch after pacman's recent update to libalpm.so.16.

The fix is basically the upgrade of some dependencies and small code fixes needed for such new versions.

#1007 
#1008


m1cham0v:
> I was only able to reproduce this issue with `firefox-nightly-bin`, although it seems like an issue with the package itself

@hbritto,
cause the problem appeared only after updating to new `libalpm`, `pacman`, `yay` and `aura` versions and appeared at all packages with `install` command inside perhaps of `${pkgdir}` variable missmatch

Issue #980, opened 2025-05-10
Feature Suggestion: Deprecate doas setting from the config file (and add an alternative)


PROStefi1000:
My suggestion is to deprecate (or just outright remove) the doas (bool) config value under [general] in the .toml configuration file. Why? I'm suggesting replacing it with an alternative setting, a value that's set to a string. That string will be the name of the privilege escalator (or path pointing directly to the executable).
My point is, there aren't just doas and sudo, there are also sudo-rs, super, vsys, sus, su1, op, calife, etc. So I think rather than adding a bagillion different bool config values, it would be easier to replace the config value with a string **now**, before the project gets even bigger, because then it would be harder to replace.
It wouldn't change any default behaviour, and would just improve the experience for those like me, who use neither sudo nor doas.

Of course, this is just a suggestion, no pressure.

fosskers:
This is a good suggestion.

PR #1010, opened 2025-12-17
remove `doas` setting, add `elevator`

dchinmay2:
fixes https://github.com/fosskers/aura/issues/980

### DIFF
--- a/rust/aura-pm/src/env.rs
+++ b/rust/aura-pm/src/env.rs
@@ -19,6 +19,7 @@ use std::ops::Not;
 use std::path::Path;
 use std::path::PathBuf;
 use unic_langid::LanguageIdentifier;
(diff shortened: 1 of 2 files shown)


rust/aura-pm/src/env.rs
```diff
@@ -226,7 +223,7 @@ impl Env {
 struct RawGeneral {
     cpus: Option<u32>,
     editor: Option<String>,
-    doas: Option<bool>,
+    elevator: Option<PathBuf>,
```
fosskers: This will unfortunately break old configs. We would need to continue to support that `doas` term as a deprecated option.

fosskers (changes requested):
A positive change in general. Please see my comment about the breaking change.

PR #1019, opened 2026-03-21
Ensure upgrade targets do not expand beyond installed packages

### LINKED ISSUE #975: `ignores` in Config File Does Not Prevent Reinstallation Prompt for Installed AUR Package in Package Base
- **Summary:** When running `aura -Au`, Aura prompts to reinstall the already installed AUR package `ttf-ms-win11` despite it being listed in the `ignores` section of the config file. The package `ttf-ms-win11` is part of the Package Base `ttf-ms-win11` (https://aur.archlinux.org/pkgbase/ttf-ms-win11), and I am concerned whether this affects how Aura handles `ignores`.  

- **Steps to Reproduce:**  
  1. Add `ignores = ["ttf-ms-win11"]` to `~/.config/aura/config.toml`.  
  2. Ensure the package `ttf-ms-win11` is already installed on the system.  
  3. Run `aura -Au` to check for updates to AUR packages.  
  4. Observe the prompt: `aura :: ttf-ms-win11 is marked 'ignored'. Install anyway? [Y/n]`.  

- **Expected Behavior:** According to the [Aura User Guide](https://fosskers.github.io/aura/), the `ignores` list should prevent Aura from updating or prompting to reinstall installed packages. Since `ttf-ms-win11` is in `ignores` and already installed, Aura should not prompt to reinstall it.  

- **Actual Behavior:** Aura still prompts to reinstall `ttf-ms-win11`. If `--noconfirm` is used, Aura automatically rebuilds and reinstalls the package, ignoring the `ignores` setting.  

- **Environment:**  
  - OS: Arch Linux  
  - Aura Version: 4.0.8
  - Package: `ttf-ms-win11` (AUR, not a dependency of any other package)  

- **Additional Notes:**  
  - The package `ttf-ms-win11` is part of the Package Base `ttf-ms-win11` (https://aur.archlinux.org/pkgbase/ttf-ms-win11). I suspect that being part of a Package Base affects how Aura handles `ignores`, especially if the Package Base contains multiple subpackages or requires updates.  
  - Similar Example: The Package Base `clion` (https://aur.archlinux.org/pkgbase/clion) contains subpackages like `clion`, `clion-jre`, etc. 

- **Request:**  
  - Please clarify if this is the intended behavior of Aura when handling packages within a Package Base.  
  - If not, consider fixing the `ignores` functionality to fully skip ignored packages, even when they are part of a Package Base with changes. 

- **Closing Note:**  
  Thank you for your time and support! If you need more information, feel free to let me know.

razor-x:
Closes https://github.com/fosskers/aura/issues/975

Hopefully I understood this correctly:

1. The initial filtered list should be the allowed package names to upgrade based only on what is installed.
2. The next two lists are for getting metadata from FAUR. (This is the part I don't fully understand).
3. Regardless of what happens in step (2), the final list should never contain new package names, so we restrict it to the original set.

Initially, I thought I could just apply the original filter again, but step (2) can still expand to more packages than was is currently installed. If I do just this,

```rust
        let names = to_upgrade
            .iter()
            .map(|(old, _)| old.name.as_ref())
            .chain(vcs.iter().map(|p| p.name.as_ref()))
            .filter(|name| env.aur.ignores.contains(*name).not());

        install(fll, &env, Mode::Upgrade, names)?;
```

then "aura :: AUR packages to upgrade:" shows the ignored package, `ttf-ms-win11`, while the later "aura :: AUR packages:" will have it filtered out. 

Since ignored packaged should not be in "AUR packages to upgrade:" I went with the approach in the PR.


razor-x:
@fosskers or @saifshahriar any feedback on this approach? 🙏🏻

fosskers:
Thanks, I will review.

PR #1024, opened 2026-07-02
feat: add option to view PKGBUILD before installing

jhossbach:
Adds `-kk` option (e.g. `aura -Akk <package>) for viewing the PKGBUILD

I tried to adhere to the existing code format but some design considerations I leave up to you @fosskers, namely
1. `u8` or some `Enum` for the `-k` option? E.g. `-kkk` silently fails now so an enum could avoid that
2. I used the `proceed` as the message, or should it be something like `Viewing the PKGBUILD`?

Happy reviewing!
P.S. Are all of the commented-out parts in the .ftl files where translations for that specific language are missing? If so I could also provide some of the missing translations for de-DE :)


fosskers:
Hi there, forgive the delay on this. 

> Are all of the commented-out parts in the .ftl files where translations for that specific language are missing? If so I could also provide some of the missing translations for de-DE :)

Correct! When "missing" like this, those fields default to English, the universal fallback.

